PR #14, opened 2019-05-03
[keycloak] Use nindent and with and improve indentation

unguiculus:
Signed-off-by: Reinhard Nägele <unguiculus@gmail.com>


unguiculus:
Will be part of #20.

PR #18, opened 2019-05-08
[keycloak] Add OpenShift route

srueg:
To support installations on OpenShift.

Signed-off-by: Simon Rüegg <simon.ruegg@vshn.ch>


charts/keycloak/templates/route.yaml
```diff
@@ -0,0 +1,35 @@
+{{- $route := .Values.keycloak.route -}}
+{{- if $route.enabled -}}
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  name: {{ template "keycloak.fullname" . }}
+{{- with $route.annotations }}
```
unguiculus: Maybe just use a `{{- with $route }}` around everything?

srueg: My idea was to copy the style of the `ingress.yaml` but if you prefer `with`, I'll change it.

unguiculus: I'm going to update all resources. See https://github.com/codecentric/helm-charts/pull/20/commits/1b6b1070edf0b3905823f703f2274d9b9cdaa914

But I guess ingress could be improved as well. We can leave it as is for now if you like.

srueg: Yes, I'd prefer to leave it as is. I'm not really a friend of `with` as it hides the global scope.

srueg:
Anything left to do here?

Issue #24, opened 2019-05-12
keycloak chart and istio


xeor:
Istio is unable to inject it's sidecar, the `securityContext` is defined directly on the statefulSet, overwriting what the sidecar needs to do, ending up giving an error like `iptables v1.6.0: can't initialize iptables table "nat": Permission denied (you must be root)`

This is discussed in https://github.com/istio/old_issues_repo/issues/316

I'm testing out replacing

```
      securityContext:
        fsGroup: 1001
        runAsNonRoot: true
        runAsUser: 1000
```

directly on the statefulSet, with

```
      securityContext:
        runAsNonRoot: true
        runAsUser: 1000
```

on the containers definition instead.

Will attach a PR on this soon if everything seams to work..

denniseffing:
There are two possible solutions:

1. As a cluster operator, use Istio CNI instead of Istio init containers ([Istio CNI documentation](https://istio.io/docs/setup/kubernetes/additional-setup/cni/))
2. Change the Keycloak Helm chart so that `securityContext` is only defined on the main container instead of defining it on the top level `spec`. Defining it on the top level `spec` results in injected containers of Istio inherting the security context configuration, thus causing this issue.

xeor:
> 1. As a cluster operator, use Istio CNI instead of Istio init containers ([Istio CNI documentation](https://istio.io/docs/setup/kubernetes/additional-setup/cni/))

Thanks for the tip.. This worked like a charm, and the "problem" gone "global". This is the `solution`

> 2. Use Istio version 1.1.3 or higher ([1.1.3 includes a fix for this](https://istio.io/about/notes/1.1.3/))

I was using `1.1.4`, in the version that failed. I upgraded to `1.1.6` and enabled cni at the same time, so I'm not sure if `1.1.6` contains a fix or not..

> 3. Change the Keycloak Helm chart so that `securityContext` is only defined on the main container instead of defining it on the top level `spec`. Defining it on the top level `spec` results in injected containers of Istio inherting the security context configuration, thus causing this issue.

A `best practice` might be to set it on the container-level, but then `fsGroup` coulnt be used. Looks like keycloak starts fine without, but not sure if there are any other bugs related.

denniseffing:
> I was using `1.1.4`, in the version that failed. I upgraded to `1.1.6` and enabled cni at the same time, so I'm not sure if `1.1.6` contains a fix or not..

[...]

PR #25, opened 2019-05-14
[keycloak] Fix iptables permission denied error on Istio 1.1.2 and below 

denniseffing:
Istio 1.1.2 and below does not include an explicit security context configuration for its init containers. They instead inherit the security context configuration defined on the PodSpecification level.

This change splits security context config into separate pod and container security contexts. This allows for a more fine grained security context configuration and is less obtrusive regarding additionally injected containers.

Fixes: #24
Signed-off-by: Dennis Effing <dennis.effing@codecentric.de>


charts/keycloak/templates/test/test-pod.yaml
```diff
@@ -13,11 +13,13 @@ metadata:
...
 spec:
   securityContext:
-{{ toYaml .Values.test.securityContext | indent 8 }}
+{{ toYaml .Values.test.securityContext | indent 4 }}
   containers:
     - name: {{ .Chart.Name }}-test
       image: "{{ .Values.test.image.repository }}:{{ .Values.test.image.tag }}"
       imagePullPolicy: {{ .Values.test.image.pullPolicy }}
+      securityContext:  
+{{ toYaml .Values.test.containerSecurityContext | indent 8 }}
```
unguiculus: Needs to be documented in readme.

PR #26, opened 2019-05-14
[keycloak] Add service account creation

denniseffing:
This PR adds service account creation to the Keycloak helm chart.

Service account creation can be configured by setting `keycloak.serviceAccount.create` and `keycloak.serviceAccount.name`.


charts/keycloak/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    app: {{ template "keycloak.name" . }}
+    chart: {{ template "keycloak.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  name: {{ template "keycloak.fullname" . }}
+{{- end -}}
```
unguiculus: Add line break at end of file.

charts/keycloak/templates/statefulset.yaml
```diff
@@ -38,6 +38,9 @@ spec:
 {{ toYaml . | indent 8 }}
 {{- end }}
       enableServiceLinks: {{ .Values.keycloak.enableServiceLinks }}
+{{- if or (.Values.keycloak.serviceAccount.create) (.Values.keycloak.serviceAccount.name) }}
```
unguiculus: The if statement is unnecessary. Please remove it.

charts/keycloak/values.yaml
```diff
@@ -28,6 +28,13 @@ keycloak:
 
   enableServiceLinks: false
 
+  serviceAccount:
+    # Specifies whether a service account should be created
+    create: true
```
unguiculus: I'd set it to `false` by default.

denniseffing:
I added the requested changes. CI seems to complain because it wants me to bump the version number again, however imo that doesn't make much sense.

PR #27, opened 2019-05-14
[keycloak] Move config checksum annotation to pod template

srueg:
The annotation should be set on the pods and not on the StatefulSet
istelf. Otherwise pods won't be restarted on config changes.

Signed-off-by: Simon Rüegg <simon.ruegg@vshn.ch>


unguiculus:
Thanks for fixing this 🎉. I just recently fixed this for another chart and missed it here.

PR #48, opened 2019-06-06
Improve NOTES.txt to avoid misleading ports being emitted, in the eve…

alsmith:
Avoid misleading ports being emitted, in the event that they have been overridden by --set.


charts/mailhog/templates/NOTES.txt
```diff
@@ -7,10 +7,10 @@ Web UI:
 =======
 
 export POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app={{ template "mailhog.name" . }},release={{ .Release.Name }}" -o jsonpath="{.items[0].metadata.name}")
-kubectl port-forward --namespace {{ .Release.Namespace }} $POD_NAME 8025
+kubectl port-forward --namespace {{ .Release.Namespace }} $POD_NAME {{ .Values.service.port.http }}
```
unguiculus: That's not correct. The ports used here are container ports which are hard-coded, not service port.

PR #20, opened 2019-05-09
[keycloak] Create new major chart release

unguiculus:
This PR collects a number of changes for a new major chart release.

- [x] Use nindent and improve indentation
- [x] Use new support for running scripts at startup
- [x] Upgrade Keycloak to 6.0.1
- [x] Make actionTokens cache configurable
- [x] Upgrade Postgresql dependency
- [x] Update labels to use new Kubernetes label conventions
- [x] Improve documentation on CLI handling
- [x] Add flag for toggling CLI execution globally
- [x] Use password file for admin user
- [x] Add upgrading section to README

Fixes: #28 
Fixes: #44 
Fixes: #45

PR #56, opened 2019-06-19
[keycloak] Add realm importer sidecar

davidholsgrove:
Inspired by [stable/grafana](https://github.com/helm/charts/tree/master/stable/grafana) use of a sidecar to import dashboards from configmaps with a specified label, this PR adds similar functionality to keycloak for importing realms by configmap with a label `keycloak_realm` for example.

**This PR includes a rebase of https://github.com/codecentric/helm-charts/pull/20 against `codecentric/helm-charts` master branch. Once that PR is closed, this PR can contain only the sidecar importer functionality**

The JBoss entrypoint startup scripts functionality is used to launch a background process by `realm-importer.sh` to monitor the volume shared in the pod between the keycloak and sidecar containers.

_The logic for adding or updating a realm uses `kcadm.sh` but this may well not be the best way to import a realm to keycloak, perhaps a wildfly cli command or the rest API would be better, but beyond my keycloak experience_


davidholsgrove:
Just saw how comprehensive the circleci tests are - pretty cool stuff :)
If we're okay with this PR in principal, a second test could be added to create a non-master realm by configmap, and test login that way too.

Also, the `k8s-sidecar` currently has a [PR](https://github.com/kiwigrid/k8s-sidecar/pull/32) open to add the same functionality to scan for kubernetes secrets, which may be a more appropriate resource to use for realm imports. If that PR is merged, the `RESOURCE` env var on the sidecar for realm-importer can be set to secret.

unguiculus:
Thanks for the PR. You should not base it on #20 which is only a draft so far and needs further work. I only looked briefly over your PR. The Keycloak chart already supports configuring additional containers (`exetraInitContainers`, `extraContainers`). So, I'm kind of hesitant to accept this as is. Anyways, I'd have to have a deeper look first.

davidholsgrove:
Thanks @unguiculus.
I initially started by using extraContainers, extraVolumes etc, but the lack of rbac on the keycloak chart made granting namespaced or cluster wide access to configmaps a bit more difficult.
Rather than forking the chart, I thought contributing (off by default) functionality which is parallel to at least how grafana does it might benefit others too.

Let me know if you have any other questions / concerns 👍

unguiculus:
Thanks again for your valuable contribution. Let me think it through. Anyways, I think we should leverage existing functionality and avoid duplication. You can already configure a service account. If you use an existing one with the required permissions, you should be able to achieve what you need. I've already thought about adding additional RBAC support which would also be necessary in case someone wants to use KUBE_PING instead of DNS_PING.

davidholsgrove:
Hi @unguiculus,

Included below is an implementation of this functionality using the GA version of the chart and the existing config value sections exposed;

```
keycloak:
  serviceAccount:
    create: true
    name: keycloak

  extraVolumes: |
    - name: realms-volume
      emptyDir: {}

  extraVolumeMounts: |
    - name: realms-volume
      mountPath: /realms

  extraEnv: |
    - name: REALMS_FOLDER
      value: "/realms"
    - name: REALM_UPDATER_INTERVAL
      value: "10"

  # Sidecar container to monitor kubernetes configmap maps for matching labels. Copy the realm json to shared volume with keycloak
  extraContainers: |
    - name: sc-realms
      image: kiwigrid/k8s-sidecar:0.0.18
      imagePullPolicy: IfNotPresent
      env:
        - name: LABEL
          value: "keycloak_realm"
        - name: FOLDER
          value: "/realms"
        # If specified, the sidecar will search for realm config-maps inside this namespace.
        # Otherwise the namespace in which the sidecar is running will be used.
        # It's also possible to specify ALL to search in all namespaces
        - name: NAMESPACE
          value: "ALL"
      volumeMounts:
        - name: realms-volume
          mountPath: /realms

  # Startup script will launch a background process which monitors the shared volume for realms.
  # If a new file is created (by the sidecar) the background process will use kcadm.sh to create / attempt to update the realm with keycloak.
  startupScripts:
    realm-importer.sh: |
      #!/bin/sh
      export PATH=/opt/jboss/keycloak/bin:$PATH
      export KEYCLOAK_URL="http://localhost:8080/auth"

      function backgroundRealmImporter() {
          while ! curl ${KEYCLOAK_URL} &>/dev/null; do
              echo "[realm-importer] Waiting on keycloak to start"
              sleep 1
          done

          while true; do
              # Process realm configmaps at default every 10 seconds
              sleep ${REALM_UPDATER_INTERVAL:-10}

              if [[ -d "${REALMS_FOLDER}" ]]; then
                  for file in ${REALMS_FOLDER}/*; do
                      if [[ "${file}" == *.json ]]; then
                          name="$(cat ${file} | jq -r .realm)"
                          echo "[realm-importer] Processing ConfigMap '${file}' (realm '${name}')"

                          echo "[realm-importer] kcadm get credentials"
                          kcadm.sh config credentials --server ${KEYCLOAK_URL} --realm master --user ${KEYCLOAK_USER} --password ${KEYCLOAK_PASSWORD}
                          rc=$?

                          if [ "${rc}" != "0" ];then
                              echo "[realm-importer] kcadm.sh failed to obtain keycloak credentials" >&2
                              continue
                          fi

                          if [ "$(kcadm.sh get realms --fields realm | jq -rS .[].realm | grep -c ${name})" == "1" ]; then
                              echo "[realm-importer] Updating realm '${name}'"
                              kcadm.sh update realms/${name} -f ${file}
                          else
                              echo "[realm-importer] Creating realm '${name}'"
                              kcadm.sh create realms -f ${file}
                          fi
                          rc=$?

                          if [ "${rc}" == "0" ];then
                              echo "[realm-importer] Import complete, moving '${file}' to '${file}.imported'"
                              mv ${file} ${file}.imported
                          else
                              echo "[realm-importer] Import failed, moving '${file}' to '${file}.failed'" >&2
                              mv ${file} ${file}.failed
                          fi
                      fi
                  done
              fi
          done
      }

      # Confirm we have required configuration before starting
      if [ -z ${!KEYCLOAK_PASSWORD} ]; then
        echo "[realm-importer] Checking KEYCLOAK_PASSWORD_FILE '${KEYCLOAK_PASSWORD_FILE}' exists"
        while ! ls ${KEYCLOAK_PASSWORD_FILE} &>/dev/null; do
            echo "[realm-importer] Waiting on KEYCLOAK_PASSWORD_FILE '${KEYCLOAK_PASSWORD_FILE}' to exist"
            sleep 1
        done
        export KEYCLOAK_PASSWORD="$(cat ${KEYCLOAK_PASSWORD_FILE})"
      fi

      missing_vars=""
      for required_var in REALMS_FOLDER KEYCLOAK_URL KEYCLOAK_USER KEYCLOAK_PASSWORD; do
          if [ -z ${!required_var} ]; then
              echo "Error - required environment var not set '${required_var}'" >&2
              missing_vars="true"
          fi
      done
      if [ "${missing_vars}" != "" ]; then exit 1; fi

      # Start background realm folder monitor and importer
      echo "[realm-importer] Folder '${REALMS_FOLDER}' will be monitored to import realms to ${KEYCLOAK_URL}"
      backgroundRealmImporter &
      pid=$!
      echo "[realm-importer] Background monitoring of ${REALMS_FOLDER} started (pid '${pid}')"
```

This relies on creating the service account as you suggested, with a (cluster)role and (cluster)rolebinding externally from the helm chart like;

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: keycloak
  namespace: keycloak
  labels:
    app: keycloak
    chart: keycloak
    heritage: keycloak
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: keycloak
subjects:
  - kind: ServiceAccount
    name: keycloak
    namespace: keycloak
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: keycloak
  namespace: keycloak
  labels:
    app: keycloak
    chart: keycloak
    heritage: keycloak
rules:
- apiGroups: [""] # "" indicates the core API group
  resources: ["configmaps"]
  verbs: ["get", "watch", "list"]
```

If you're not looking to add similar functionality out-of-the-box to the keycloak helm chart, I believe we can close this PR

thanks
David

swistaczek:
It would be awesome if we can build such functionality to this helm chart 💘.

unguiculus:
Closing this as suggested by @davidholsgrove. Please open a new PR if you need anything to be added.

PR #67, opened 2019-07-07
Fix keycloak pods label filtering

geeckmc:
The current version introduces an error because the pods do not have the requested label


unguiculus:
Good catch. Thanks. I must have missed that. Please bump the patch version of the chart and fix DCO so this can be merged.

See https://github.com/codecentric/helm-charts/pull/67/checks?check_run_id=163358541

PR #92, opened 2019-08-05
[stable/keycloak] Make JGroups discovery configuration configurable

swistaczek:
Hey 👋 ,
Thanks for keeping this chart up-to-date and extending features! There is one missing feature which I would like to use for my deployment. However, I've already found a solution by maintainer @unguiculus - https://github.com/helm/charts/pull/12229/commits . I have ported this solution to the latest master version. It would be great if we can get this change into upstream.

Have a great day!

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,5 +1,5 @@
 name: keycloak
-version: 5.1.6
+version: 5.2.0
 appVersion: 6.0.1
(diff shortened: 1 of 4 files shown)


unguiculus:
Thanks for the PR. I need to think this through again since I had withdrawn my PR you're referring to.

unguiculus:
For some reason CircleCI doesn't like this PR, so I cherry-picked your commit to #122.

PR #97, opened 2019-08-08
tls hosts should be quoted

pluqueTheLuxe:
they should be quoted to avoid issues on wildcard characters

actually we have issue with our wildcard chars: error parsing template.yml: error converting YAML to JSON: yaml: line 29: did not find expected alphabetic or numeric character

current output
```
spec:
  tls:
    - hosts:
        - *.development.example.com
```

desired output
```
spec:
  tls:
    - hosts:
        - '*.development.example.com'
```

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,5 +1,5 @@
 name: keycloak
-version: 5.1.5
+version: 5.1.6
 appVersion: 6.0.1
 description: Open Source Identity and Access Management For Modern Applications and Services
 keywords:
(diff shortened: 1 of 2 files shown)


unguiculus:
Thanks for the PR. Please bump the patch version of the chart and and fix DCO. See https://github.com/codecentric/helm-charts/pull/97/checks?check_run_id=189015426

pluqueTheLuxe:
> Thanks for the PR. Please bump the patch version of the chart and and fix DCO. See https://github.com/codecentric/helm-charts/pull/97/checks?check_run_id=189015426

it should be ok now

PR #119, opened 2019-10-04
Make securityContext configurable for mailhog

ccremer:
Making the security context configurable makes mailhog able to run on Openshift, where the security context gets applied automatically (and actually throws errors like these: `1000 is not an allowed group spec.containers[0].securityContext.securityContext.runAsUser: Invalid value: 1000: must be in the ranges: [1003220000, 1003229999]]`)

The default values are unchanged as to not break existing deployments.


ccremer:
Thank you very much :)

PR #120, opened 2019-10-07
[keycloak] Bump keycloak to 7.0.0

mxandeco:
There is a new keycloak version out there, this pr is a simple version bump, have tried a fresh install and everything looked fine. 

Not sure if upgrades will be straight forward or if i missed something.

#121

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 name: keycloak
-version: 5.1.7
-appVersion: 6.0.1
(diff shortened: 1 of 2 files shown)


unguiculus:
Thanks for the PR. The update is included in a number of changes for a new major release. See #122.

mxandeco:
> Thanks for the PR. The update is included in a number of changes for a new major release. See #122.

thanks @unguiculus is any ETA for that pr to be merged? thanks.

PR #123, opened 2019-10-15
[mailhog] Fix default hosts values for Ingress and their description in README

tyranron:
To save some debugging time.


unguiculus:
Thanks. Please fix DCO so this can get merged.

tyranron:
@unguiculus oh, didn't notice that. Done!

PR #128, opened 2019-10-25
[keycloak] Document name overrides in values file

gs11:
Signed-off-by: gs11 <gustav.sinder@gmail.com>


gs11:
I guess `lint-charts` failed due to the chart version being unchanged, should I simply increment the patch version?

unguiculus:
Thanks. While you're at it, please document `nameOverride` as well and bump the patch version.

gs11:
Made a try but I'm not sure why the naming logic looks like it does @unguiculus

PR #132, opened 2019-10-30
[keycloak] Add support for reading DB_USER from a k8s Secret

gzur:
Currently, only `DB_PASSWORD` can be loaded from a Secret, but not `DB_USER`.

We are using an operator pattern for db credential provisioning (think Hashicorp Vault) - and having a static username is not compatible with this pattern, since you are creating and deleting database roles as consuming pods go up and down.

This PR adds support for (optionally) reading the DB user from a secret.

The previous implementation used a variable named `keycloak.persistence.existingSecretKey` - which makes sense as long as you are only reading one value from the Secret. 

By adding another value to read from the Secret, this naming scheme is rendered unclear, so  `keycloak.persistence.existingSecretKey` becomes `keycloak.persistence.existingSecretPasswordKey` - which breaks backwards-compatibility, thus the **Major** version bump to `7.0.0`

### DIFF
--- a/charts/keycloak/README.md
+++ b/charts/keycloak/README.md
@@ -380,14 +380,15 @@ The headless service that governs the StatefulSet is used for DNS discovery.
 ## Upgrading
 
 ### From chart versions < 7.0.0
-Version 7.0.0 is a minor update - but it does break backwards-compatibility with the existing 
+Version 7.0.0 is a minor update - but it does break backwards-compatibility with the existing
 `keycloak.persistence.existingSecret` scheme.
-#### Changes in Configuring Database Credentials from an Existing Secret 
+
+#### Changes in Configuring Database Credentials from an Existing Secret
 
 Now both `DB_USER` and `DB_PASS` can be read from an existing Kubernetes Secret. This is a requirement if you are
 provisioning database credentials dynamically - either via an Operator or some secret-management engine.
 
-The variable referencing the Password Key Name has been renamed from `keycloak.persistence.existingSecretKey` 
+The variable referencing the Password Key Name has been renamed from `keycloak.persistence.existingSecretKey`
 to `keycloak.persistence.existingSecretPasswordKey`
 
 A new, optional variable for referencing the Username Key Name for populating the `DB_USER` env has been added:
@@ -395,7 +396,7 @@ A new, optional variable for referencing the Username Key Name for populating th
 
 If left unset, `DB_USER` will be populated by the `dbUser` Helm variable.
 
-###### Example configuration: 
+###### Example configuration:
 ```yaml
 keycloak:
   persistence:
--- a/charts/keycloak/README.md
+++ b/charts/keycloak/README.md
@@ -111,7 +111,8 @@ Parameter | Description | Default
 `keycloak.route.tls.termination` | TLS termination of the route. Can be `edge`, `passthrough` or `reencrypt` | `edge`
 `keycloak.persistence.deployPostgres` | If true, the PostgreSQL chart is installed | `false`
 `keycloak.persistence.existingSecret` | Name of an existing secret to be used for the database password (if `keycloak.persistence.deployPostgres=false`). Otherwise a new secret is created | `""`
-`keycloak.persistence.existingSecretKey` | The key for the database password in the existing secret (if `keycloak.persistence.deployPostgres=false`) | `password`
+`keycloak.persistence.existingSecretPasswordKey` | The key for the database password in the existing secret (if `keycloak.persistence.deployPostgres=false` and `keycloak.persistence.existingSecret != ""`) | `""`
+`keycloak.persistence.existingSecretUsernameKey` | The key for the database username in the existing secret (if `keycloak.persistence.deployPostgres=false` and `keycloak.persistence.existingSecret != ""`). Will default to the value of `.keycloak.persistence.dbUser` if left unset. | `""`
 `keycloak.persistence.dbVendor` | One of `h2`, `postgres`, `mysql`, or `mariadb` (if `deployPostgres=false`) | `h2`
 `keycloak.persistence.dbName` | The name of the database to connect to (if `deployPostgres=false`) | `keycloak`
 `keycloak.persistence.dbHost` | The database host name (if `deployPostgres=false`) | `mykeycloak`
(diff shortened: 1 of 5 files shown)


gzur:
There is now another PR (https://github.com/codecentric/helm-charts/pull/135) for this exact same functionality opened by @liamdawson - the most pertinent differences are:

- https://github.com/codecentric/helm-charts/pull/135) maintains backwards-compatibility (and thus evades the Major version bump) by only introducing new variables - leaving `keycloak.persistence.existingSecretKey` as it is - whereas this PR renames that variable to `existingSecretPasswordKey` for clarity - which in my view is more explicit and thus preferred.
Though I obviously have a preference regarding the variable naming, I leave it to the maintainers to balance the Major version bump against some minor semantic clarity.

- https://github.com/codecentric/helm-charts/pull/135) also gives you the freedom to read the Username from another Secret than the Password - which is definitely more dynamic, and I wish I'd thought of it :) - but I can update this PR to include this functionality.

liamdawson:
I closed my PR as I'm not giving it the attention to keep it up to date (I stopped using Helm)

charts/keycloak/README.md
```diff
@@ -377,6 +379,31 @@ The headless service that governs the StatefulSet is used for DNS discovery.
 
 ## Upgrading
 
+### From chart versions < 7.0.0
+Version 7.0.0 is a minor update - but it does break backwards-compatibility with the existing 
+`keycloak.persistence.existingSecret` scheme.
```
unguiculus: Add an empty line.

charts/keycloak/templates/secret-db.yaml
```diff
@@ -8,4 +8,7 @@ metadata:
 type: Opaque
 data:
   {{ include "keycloak.dbPasswordKey" . }}: {{ .Values.keycloak.persistence.dbPassword | b64enc | quote }}
+{{- if .Values.keycloak.persistence.existingSecretUsernameKey }}
```
unguiculus: Since this is a breaking change anyways, I'd prefer it if we handled username and password the same and always put both into the secret.

gzur: Done.

charts/keycloak/templates/_helpers.tpl
```diff
@@ -165,7 +176,17 @@ Create environment variables for database configuration.
...
+{{- /*
+Load the dbUser from either a Secret or directly from .Values if existingSecretUsernameKey is not set.
+*/ -}}
+{{- if and .Values.keycloak.persistence.existingSecret .Values.keycloak.persistence.existingSecretUsernameKey }}
+  valueFrom:
+    secretKeyRef:
+      name: {{ include "keycloak.externalDbSecret" . }}
+      key: {{ include "keycloak.dbUserKey" . | quote }}
+{{- else }}
+  value: {{ .Values.keycloak.persistence.dbUser }}
```
unguiculus: Only read from secret

gzur:
How does it look now?

charts/keycloak/README.md
```diff
@@ -377,6 +379,33 @@ The headless service that governs the StatefulSet is used for DNS discovery.
 
 ## Upgrading
 
+### From chart versions < 7.0.0
```
unguiculus: Add empty line

charts/keycloak/README.md
```diff
@@ -377,6 +379,33 @@ The headless service that governs the StatefulSet is used for DNS discovery.
...
+Version 7.0.0 is a minor update - but it does break backwards-compatibility with the existing
```
unguiculus: I wouldn't call it a minor update. We do bump the major version. Just leave that out.

charts/keycloak/README.md
```diff
@@ -377,6 +379,33 @@ The headless service that governs the StatefulSet is used for DNS discovery.
...
+`keycloak.persistence.existingSecret` scheme.
+
+#### Changes in Configuring Database Credentials from an Existing Secret
+
+Both `DB_USER` and `DB_PASS` are always read from a Kubernetes Secret. This is a requirement if you are
+provisioning database credentials dynamically - either via an Operator or some secret-management engine.
+
+The variable referencing the Password Key Name has been renamed from `keycloak.persistence.existingSecretKey`
```
unguiculus:
```suggestion
The variable referencing the password key name has been renamed from `keycloak.persistence.existingSecretKey`
```

charts/keycloak/README.md
```diff
@@ -377,6 +379,33 @@ The headless service that governs the StatefulSet is used for DNS discovery.
...
+to `keycloak.persistence.existingSecretPasswordKey`
+
+A new, optional variable for referencing the Username Key Name for populating the `DB_USER` env has been added:
```
unguiculus:
```suggestion
A new, optional variable for referencing the username key name for populating the `DB_USER` env has been added:
```

charts/keycloak/README.md
```diff
@@ -377,6 +379,33 @@ The headless service that governs the StatefulSet is used for DNS discovery.
...
+`keycloak.persistence.existingSecretUsernameKey`.
+
+If `keycloak.persistence.existingSecret` is left unset, a new Secret will be provisioned populated with the `dbUser`
+and `dbPassword`  Helm variables.
```
unguiculus: Don't wrap the line here. One sentence per line, please.

gzur:
I don't understand[ what's going on with DCO.](https://github.com/codecentric/helm-charts/pull/132/checks?check_run_id=420222872)
The commit looks signed to me.

It was created via github's Change Suggestion mechanism.

gzur:
> I don't understand[ what's going on with DCO.](https://github.com/codecentric/helm-charts/pull/132/checks?check_run_id=420222872)
> The commit looks signed to me.
> 
> It was created via github's Change Suggestion mechanism.

Nevermind, I managed to solve it through furious force-pushing.

charts/keycloak/templates/_helpers.tpl
```diff
@@ -125,12 +125,23 @@ Create the name for the database password secret key.
 */}}
 {{- define "keycloak.dbPasswordKey" -}}
 {{- if .Values.keycloak.persistence.existingSecret -}}
-  {{- .Values.keycloak.persistence.existingSecretKey -}}
+  {{- required "missing existingSecretPasswordKey" .Values.keycloak.persistence.existingSecretPasswordKey -}}
```
unguiculus: Why do you use different logic for username and password key?

gzur: No good reason. Those definitions now use the same logic.

gzur: No good reason. Those definitions now use the same logic.

One can now be configured without the other (which will fall back to  a default). 
This is more flexible than forcing the user to set both or neither.

unguiculus:
Thanks!

PR #142, opened 2019-11-25
Move ingress to networking.k8s api

danylohetmantsev:
Accordingly to latest kubernetes docs https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource

Ingress right now moved outside of the extensions and right now it's part for the networking.k8s.io/v1beta1 api.

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 6.0.3
+version: 6.0.4
 appVersion: 7.0.0
 description: Open Source Identity and Access Management For Modern Applications and Services
 keywords:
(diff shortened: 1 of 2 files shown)


Flowkap:
Please ... I just wanted to create PR and found theres one already ^^

unguiculus:
There's not rush to do this. `extensions/v1beta1` will be available till K8s 1.20, while `networking.k8s.io/v1beta1` is only available from 1.14.

unguiculus:
I think we can go ahead and get this in now. Please fix DCO and resolve conflicts.

tyranron:
ping @danylohetmantsev @unguiculus 

Would you be so kind to make it finished? In [1.18 it's disabled already](https://github.com/kubernetes/kubernetes/blob/0f1a27bceeca8f1d37cd69f33d642e38ab414192/CHANGELOG/CHANGELOG-1.18.md#kube-apiserver-1).

code-chris:
@unguiculus 
As @tyranron said, 1.18 is released now and the "old" apigroups are disabled by default since 1.16, if I'm correct...

unguiculus:
Superseded by #219.

PR #146, opened 2019-11-27
fix(keycloak): Quote release names used in labels

### LINKED ISSUE #145: [keycloak] Chart release names aren't always strings
We have an issue using your chart (and many other public charts 😞) because our helm release names are actually numbers (IDs to be precise). Changing this would be a royal pain, and we're trying to avoid forking public charts.

Would you accept an MR to append a `| quote` to places where `.Release.Name` is being used where a string is required? i.e. labels. If I understand correctly, this should also be a completely non-breaking change for everyone else.

Thoughts?

shousper:
Closes #145

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 6.0.3
+version: 6.0.4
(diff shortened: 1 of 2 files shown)


unguiculus:
Thanks.

PR #158, opened 2020-01-07
Introduced variables to allow injection of passwords and users as files

etiennejournet:
- dbPassword_file
- dbUser_file
- password_file
- To Be Discussed: Added a sleep in launch command because I inject informations
with vault-injector and can't use init container because of istio. It's clearly not the best (but it's the most simple) so let's discuss it

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 6.1.0
+version: 6.2.0
 appVersion: 8.0.1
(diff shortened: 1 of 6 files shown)


etiennejournet:
i just removed the random sleep function, can you review please ?

etiennejournet:
can you review ?

sereinity:
Up, we would like to use this to inject secrets via vault.

Thanks

unguiculus:
Thanks for the PR. A major rewrite of the chart has been released which should support what you need.

PR #58, opened 2019-06-20
[keycloak] Add Prometheus Operator support

cablespaghetti:
Adds prometheus operator support for plugins such as https://github.com/aerogear/keycloak-metrics-spi

PR #160, opened 2020-01-10
[keycloak] add support for loading prometheus monitoring at startup

Gottox:
This PR adds an option to load the [keycloak-metrics-spi](https://github.com/aerogear/keycloak-metrics-spi) plugin at startup. This adds support for prometheus monitoring with a simple config option.

ToDo:
* [x] add documentation
* [x] increas version in Chart.yaml


unguiculus:
Thanks for the PR. The chart is generic enough to achieve what you want. Please note that #58 was merged which builds upon the chart's generic capabilities.

PR #155, opened 2020-01-02
Fix misleading information in README

mikhail-putilov:
Some time ago (I honestly don't know when), `PROXY_ADDRESS_FORWARDING` was removed from the default environment variables but wasn't removed from default values in the README. This fix resolves the mismatch.

Previously I created an [issue 138](https://github.com/codecentric/helm-charts/issues/138) regarding this breaking change that I noticed when tried to upgrade the version of the chart.

PR #161, opened 2020-01-11
[keycloak] Fix misleading information in README

Mikhail-Putilov-at-adviqo-com:
That's reopened PR: https://github.com/codecentric/helm-charts/pull/155
I had to close the previous one because I did it under my personal account.

Some time ago (I honestly don't know when), `PROXY_ADDRESS_FORWARDING` was removed from the default environment variables but wasn't removed from default values in the README. This fix resolves the mismatch.

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 6.1.0
+version: 6.1.1
 appVersion: 8.0.1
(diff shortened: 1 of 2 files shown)


suheb:
Ah! The `values.yaml` still has `PROXY_ADDRESS_FORWARDING` as the default environment variable: https://github.com/codecentric/helm-charts/blob/master/charts/keycloak/values.yaml#L99
Or am I missing something really obvious here?
@Mikhail-Putilov-at-adviqo-com @unguiculus

PR #164, opened 2020-01-20
[keycloak] Use password file instead of env var for db password

zeph:
any other solution using env variables will cause the bash to interpret the password and possibly fail

### DIFF
--- a/charts/keycloak/templates/_helpers.tpl
+++ b/charts/keycloak/templates/_helpers.tpl
@@ -79,7 +79,7 @@ Create a default fully qualified app name for the postgres requirement.
 {{/*
 Create the name for the Keycloak secret.
 */}}
-{{- define "keycloak.secretName" -}}
+{{- define "keycloak.secret" -}}
 {{- if .Values.keycloak.existingSecret -}}
   {{- tpl .Values.keycloak.existingSecret $ -}}
 {{- else -}}
@@ -90,15 +90,11 @@ Create the name for the Keycloak secret.
 {{/*
 Create the name for the database secret.
 */}}
-{{- define "keycloak.dbSecretName" -}}
-{{- if .Values.keycloak.persistence.deployPostgres }}
-  {{- include "keycloak.postgresql.fullname" . -}}
-{{- else }}
-  {{- if .Values.keycloak.persistence.existingSecret -}}
-    {{- tpl .Values.keycloak.persistence.existingSecret $ -}}
-  {{- else -}}
-    {{- include "keycloak.fullname" . -}}-db
-  {{- end -}}
+{{- define "keycloak.externalDbSecret" -}}
+{{- if .Values.keycloak.persistence.existingSecret -}}
+  {{- tpl .Values.keycloak.persistence.existingSecret $ -}}
+{{- else -}}
+  {{- include "keycloak.fullname" . -}}-db
 {{- end -}}
 {{- end -}}
 
@@ -128,21 +124,10 @@ Create the name for the password secret key.
 Create the name for the database password secret key.
 */}}
 {{- define "keycloak.dbPasswordKey" -}}
-{{- if and .Values.keycloak.persistence.existingSecret .Values.keycloak.persistence.existingSecretPasswordKey -}}
-  {{- .Values.keycloak.persistence.existingSecretPasswordKey -}}
+{{- if .Values.keycloak.persistence.existingSecret -}}
+  {{- .Values.keycloak.persistence.existingSecretKey -}}
 {{- else -}}
   password
-{{- end -}}
-{{- end -}}
-
-{{/*
-Create the name for the database password secret key - if it is defined.
-*/}}
-{{- define "keycloak.dbUserKey" -}}
-{{- if and .Values.keycloak.persistence.existingSecret .Values.keycloak.persistence.existingSecretUsernameKey -}}
-  {{- .Values.keycloak.persistence.existingSecretUsernameKey -}}
-{{- else -}}
-  username
 {{- end -}}
 {{- end -}}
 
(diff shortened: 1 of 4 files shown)


zeph:
@unguiculus @thomasdarimont  anything I can help out with to push this through?

charts/keycloak/templates/_helpers.tpl
```diff
@@ -166,11 +166,16 @@ Create environment variables for database configuration.
   value: {{ .Values.keycloak.persistence.dbName | quote }}
 - name: DB_USER
   value: {{ .Values.keycloak.persistence.dbUser | quote }}
+{{- if .Values.keycloak.persistence.existingSecret}}
+- name: DB_PASSWORD_FILE
+  value: /secrets/persistence
+{{- else }}
```
unguiculus: I'd suggest we completely move to using password files only. This is already the case for the Keycloak password.

zeph: I don't know if anybody wants to still use only the env variable, the docker container allows people to do so... I tried to be backward compatible

unguiculus: I don't think we will have a compatibilty issue here. I think an upgrade should not be an issue.

zeph: your call... I don't have visibility on all possible hacks done by people on their systems out there

zeph:
@unguiculus what changes shall I address?

zeph:
@unguiculus would u mind solve or answer the CR u opened? 
...then I'll look at these conflicts that just pop up

zeph:
hi @unguiculus , hope u are better... is there any follow up on this issue?

unguiculus:
@zeph Sorry for the delay. Not sure what I need to clarify. As mentioned in another comment, I'd like you to switch to password files completely instead of making this optional.

zeph:
@unguiculus oh, ok... so... shall I modify the code? 
u wanna take the time to? how shall we move on?

unguiculus:
Ah, I thought you could do it, but if you want me to do it I might be able to spend some time on it today.

zeph:
@unguiculus if you don't mind... at least u get the way u have it in mind

unguiculus:
OK, started looking into it.

unguiculus:
@zeph Please have a look and test it.

zeph:
@unguiculus I'm on it right now, thanks for the contribution

zeph:
@unguiculus I believe we should not put the username as part of the Secret, there is no need of encrypting such thing... it is just adding an unnecessary level of complexity

unguiculus:
@zeph The username for the database is already in the secret. There were PRs for this a while ago. I think that's ok. Anyways, this is not related to this PR. However, it's a little inconsistent now because only the DB user is in the secret wheras the Keycloak user is not.

zeph:
@unguiculus sorry, it doesn't make any sense to have the username in secret. 
I reverted your changes as they ended up in my fork (interesting behaviour of github).
...I don't get why you see it inconsistent, Secret is not meant to be used for non-sensitive information. I'm using `SealedSecrets` for that, it would mean I have to have an sealed version of non-required-to-be-encrypted value in each `values` file I have (one per environment)

Would you mind keeping the pull request minimal and just merge it?

zeph:
@thomasdarimont @lenzenowski can you please have a look at
 this pull request? seems we reached a dead end in the discussion

zeph:
@gzur I see the DB_USER is in the Secret as u proposed it... what's the added value?

gzur:
@zeph 
> @gzur I see the DB_USER is in the Secret as u proposed it... what's the added value?

In short, it's so that Keycloak can rely on credentials provisioned dynamically by an operator.

More info:  https://github.com/codecentric/helm-charts/pull/132

zeph:
@gzur do we agree that a username doesn't require to be encrypted, right?

gzur:
> @gzur do we agree that a username doesn't require to be encrypted, right?

What is the value of **not** storing the Username in a Secret?

zeph:
@gzur avoiding the need of having one encrypted value per cluster if using a proper tool like SealedSecrets

KISS (Keep It Simple and Stupid) is the base principle we shall use to work. There is absolutely NO need to put the DB_USER into the Secret element, aside of obfuscating it further and preventing people to use secret management tools

zeph:
@gzur do we agree on the #kiss principle?

gzur:
You seem to be describing a workflow I'm unfamiliar with because decoupling the Username from the Password (and thus having separate delivery mechanisms for each part of the credential pair) doesn't sound very KISS to me, and actually violates The Principle of Least Astonishment, from my point of view. 

Again, I am unfamiliar with the workflow you describe, and I have no insight into how well it gels with the larger Helm ecosystem, but I get the feeling that you are letting some rather specific configuration pain introduced by your upstream toolchain trickle into this downstream repository.

zeph:
@gzur I'm not sure if this TPLA is made up or you are serious. Helm ecosystem allows me to reference the codecentric helm repository in a requirements.yaml file, right? Good, I did that. Sounds pretty Helm ecosystem standard to me. Now, in my helm chart, I do have a template for a SealedSecret resource, I keep the value for the password in its environment yaml file.

Now, encrypting the username seems a practice you alone love. Please point me where anybody else has been doing it, as I'm pretty not aware of such practice, sounds quite over-engineering to me.

gzur:
I have no great love -nor even an opinion - on encrypting usernames since Kubernetes Secrets do not imply encryption in any way.

I implemented https://github.com/codecentric/helm-charts/pull/132 to meet needs that I and others using Vault and other operator based credential provisioning/management systems had.

If my contribution is not to your liking, you are free to make PR to revert it.

Best of luck with your pull request.

zeph:
@gzur I'm trying to understand your reason, before getting into a fight pushing a pull request that you are already bullying... I don't see the point, at all, to encrypt a username. From what you last stated seems you are using Vault also for the username, is such? Are you using it to automatically rotate along with the password?

Before we had a clear interface, an existing Secret resource with a password. Now we have a different interface, with both in... feels to me you are the one that broke the common interface for everyone else

unguiculus:
I'm going to close this PR now. I won't merge it as is. @zeph The way you are acting here doesn't lead anywhere. Please watch the tone in your words and study our code of conduct: https://github.com/codecentric/helm-charts/blob/master/CODE_OF_CONDUCT.md

PR #172, opened 2020-02-13
LoadBalancerIP to static

takaahii:


charts/keycloak/templates/service-http.yaml
```diff
@@ -14,6 +14,9 @@ metadata:
     {{- end }}
     service: http
 spec:
+{{- if $service.loadBalancerIP }}
```
unguiculus:
```suggestion
{{- with $service.loadBalancerIP }}
```

charts/keycloak/templates/service-http.yaml
```diff
@@ -14,6 +14,9 @@ metadata:
...
+  loadBalancerIP: "{{ $service.loadBalancerIP }}"
```
unguiculus:
```suggestion
  loadBalancerIP: "{{ . }}"
```

unguiculus:
Already supported in latest version.

PR #173, opened 2020-02-13
[keycloak] Allow templating for service annotation values

mackoko:
Keycloak service annotation update. Values are now passed through the `tpl` function.


mackoko:
@unguiculus @thomasdarimont  anything I can do/help with
 to push this through?

PR #175, opened 2020-02-20
[keycloak] Add parameter to override termination grace period

shousper:
Primarily useful for large clusters with a lot of Infinispan cache to rebalance on deployment.

Happy to create an issue for this if required 👌

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 7.2.0
+version: 7.2.1
 appVersion: 8.0.1
 description: Open Source Identity and Access Management For Modern Applications and Services
 keywords:
--- a/charts/keycloak/values.yaml
+++ b/charts/keycloak/values.yaml
@@ -74,6 +74,9 @@ keycloak:
     #   exec:
     #     command: ["/bin/sh", "-c", "ls"]
 
+  ## Override the default for the Keycloak container, e.g. for clusters with large cache that requires rebalancing.
+  terminationGracePeriodSeconds: 60
+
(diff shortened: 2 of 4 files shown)


charts/keycloak/values.yaml
```diff
@@ -74,6 +74,9 @@ keycloak:
     #   exec:
     #     command: ["/bin/sh", "-c", "ls"]
 
+  ## Override the default for the Keycloak container, e.g. for clusters with large cache that requires rebalancing.
+  terminationGracePeriodSeconds: 60
```
unguiculus: We should stick with the default of 30.

charts/keycloak/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 7.2.0
+version: 7.2.1
```
unguiculus: Bump minor version.

charts/keycloak/values.yaml
shousper: I believe changing the default value would be a breaking change that warrants a major version bump and a note in the upgrade guide to warn users of the change. Are you happy for me to do this as well?

If you don't mind my asking, are you running a cluster with the defaults in this chart and don't have any issues rolling the cluster? e.g. infinispan replication errors, loss of sessions and/or users getting logged out?

robsonpeixoto: But it is not changing the default value, only add the possibility to use another value.
IMHO it is not a breaking change, because `terminationGracePeriodSeconds` will keep as `60` for users that are not changing it.

unguiculus: I figured we already had 60 before. All good. 😉

PR #187, opened 2020-03-04
add labels to headless service template similar to http service

yk12OT:


yk12OT:
closing without merging - not needed (opened a PR against opendoor branch rather than master)

PR #190, opened 2020-03-11
[jenkins] Add environment variables to the init container and set revisonHistoryLimit

sagiru:
Hi there,

we are running a kubernetes cluster behind a proxy and then the init container can't pull plugins after the deployment. To enable the init container to pull stuff from outside i reuse the extraEnv vars for the init container too. 

The second patch sets the revisionHistory limit for the deployment. Sometimes we have the need to configure more or less than the default of 10 old sets.

Regards
Sascha

### DIFF
--- a/charts/jenkins/Chart.yaml
+++ b/charts/jenkins/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: jenkins
 description: The leading open source automation server
-version: 1.5.1
+version: 1.6.0
 appVersion: 2.190.1
 home: https://jenkins.io/
 icon: https://wiki.jenkins-ci.org/download/attachments/2916393/logo.png
--- a/charts/jenkins/templates/deployment.yaml
+++ b/charts/jenkins/templates/deployment.yaml
@@ -8,6 +8,7 @@ spec:
   selector:
     matchLabels:
       {{- include "jenkins.selectorLabels" . | nindent 6 }}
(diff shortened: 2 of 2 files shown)


sagiru:
Hi,

does it make sense to check a version bump for a pull request?

Regards

unguiculus:
@sagiru Every PR needs a version bump because it results in a new release when it is merged.

charts/jenkins/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: jenkins
 description: The leading open source automation server
-version: 1.5.1
+version: 1.5.2
```
unguiculus, line 4: Please bump the minor version to follow semver.

sagiru: Hi,

sry, i don't understand whats wrong with the 1.5.2?

Regards
Sascha

unguiculus: Following semver, a new feature requires a minor version bump.

sagiru: Ah, sry so a 1.6.1 would be correct.

unguiculus: 1.6.0 would be the next minor version.

sagiru: oh man, i'm sorry...maybe i should drink more coffee before i do things like that...

PR #199, opened 2020-03-26
[mailhog] Use networking.k8s.io/v1beta1 API for Ingress

tyranron:
This PR switches `mailhog`'s Ingress `apiVersion` to `networking.k8s.io/v1beta1`

## Motivation

In [1.18 Kubernetes `extensions/v1beta1` for Ingress is disabled completely](https://github.com/kubernetes/kubernetes/blob/0f1a27bceeca8f1d37cd69f33d642e38ab414192/CHANGELOG/CHANGELOG-1.18.md#kube-apiserver-1).

### DIFF
--- a/charts/mailhog/templates/ingress.yaml
+++ b/charts/mailhog/templates/ingress.yaml
@@ -1,6 +1,6 @@
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "mailhog.fullname" . -}}
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
 kind: Ingress
 metadata:
   name: {{ $fullName }}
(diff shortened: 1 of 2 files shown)


tyranron:
ping @unguiculus

charts/mailhog/templates/ingress.yaml
```diff
@@ -1,6 +1,6 @@
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "mailhog.fullname" . -}}
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
```
unguiculus, line 3: Should be v1beta1

```suggestion
apiVersion: networking.k8s.io/v1beta1
```

tyranron: meh... thanks!

PR #206, opened 2020-04-09
Mailhog replicas

bartlettt:
Added replica count configuration to mailhog


Jancis:
Any chance to resurrect this PR? I stumbled precisely into this issue - my use case is that there's a downscaler cronjob, setting `spec.replicas` to `0` for selected deployments including mailhog, but the next helm deployment never removes the `replicas: 0` and so mailhog always stays in suspended state. If the chart template would define spec.replicas, it would set the default value and wake up the deployment.

bartlettt:
Honestly I don't even remember writing this PR haha. My changes look simple enough, so feel free to open a new PR and use what I had

PR #215, opened 2020-04-30
adding X.509 Client Certificate User Authentication

chrishardis:
Signed-off-by: Christian Senet <Christian.Senet@hardis.fr>

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 7.7.0
+version: 7.7.1
 appVersion: 9.0.2
 description: Open Source Identity and Access Management For Modern Applications and Services
 keywords:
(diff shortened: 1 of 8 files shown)


chrishardis:
Thank you for your feedback.
According your answer I better understood I have miss that extraVolumeMounts and extraVolume could help to resolve the x509 certs location, therefore the configmap use both for wildfly configuration and nginx reverse proxy secret is not mandatory in the chart.
I'm going to do the same one the regular chart, I will commit a dedicated x509 setup guide if it can help.

chrishardis:
I have almost be able to do the x509 authentication on current template. The only missing part is in ingress.yaml. The backport is always http, whereas we must be in https for x509 back-end authentication.

``` yaml
  rules:
  {{- range $ingress.hosts }}
    - host: {{ . }}
      http:
        paths:
          - path: {{ $ingress.path }}
            backend:
              serviceName: {{ include "keycloak.fullname" $ }}-http
              servicePort: http
  {{- end }}
```
I propose to add in values.yaml and update ingress.yaml
``` yaml
  ingress:
    enabled: true
    path: /
    backendssl: true
```    
Any suggestion ?

unguiculus:
I guess this should work for you: #231

unguiculus:
Closing as solved.

PR #216, opened 2020-04-30
Release name in deployment labels

Jancis:
Most helm charts add `.Release.Name` to deployment labels, which makes it easier to select the resource.


unguiculus:
The chart follows Kubernetes label conventions. 

https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

Helm also follows these conventions when you create a new chart with `helm create`. In other words, the `release` label is no longer used. Instead, there is  `app.kubernetes.io/instance` which should give you what you need.

PR #236, opened 2020-05-30
[keycloak] Add comment on proxy-buffer-size for ingress-nginx

ricardoamedeiros:
Resolve bug response 502 of nginx kubernetes.

Detail:

https://www.ibm.com/support/pages/502-error-ingress-keycloak-response


ricardoamedeiros:
> Thanks. Please fix DCO and bump the patch version of the chart so this can be merged.

ready.

cablespaghetti:
@ricardoamedeiros You are a hero thanks! I never would have found this otherwise. :smile:

PR #249, opened 2020-07-10
Added configurability for startCommand & prometheus port

GregB-cc:


unguiculus:
Thanks for the PR. A major rewrite of the chart has been released which supports what you need.

PR #250, opened 2020-07-10
[keycloak] Refactor chart and upgrade to 11.0.0

unguiculus:
With this change, the Keycloak chart receives a major facelift. Opinionated stuff and things that are now baked into Keycloak's Docker image are removed. Configuration is more generic making it easier to use custom Docker images that are configured differently than the official one.


charts/keycloak/templates/test/configmap-test.yaml
```diff
@@ -4,7 +4,10 @@ kind: ConfigMap
 metadata:
   name: {{ include "keycloak.fullname" . }}-test
   labels:
-    {{- include "keycloak.commonLabels" . | nindent 4 }}
+    {{- include "keycloak.labels" . | nindent 4 }}
+  annotations:
+    helm.sh/hook: test
+    helm.sh/hook-delete-policy: hook-succeeded
```
scottrigby, line 10: 👏 👏

charts/keycloak/README.md
```diff
@@ -24,217 +24,289 @@ By default, the PostgreSQL chart requires PV support on underlying infrastructur
...
+    value: mydb
+
+extraEnvFrom: |
+  - secretRef:
+      name: '{{ include "keycloak.fullname" . }}-db'
+
+secrets:
+  - nameSuffix: db
+    stringData:
+      DB_USER: '{{ .Values.dbUser }}'
```
scottrigby, line 243: Wouldn't secrets also have to be a yaml string literal (`|`) in order to pass templating like this (also example below `my-test-annotation: Test secret for {{ include "keycloak.fullname" . }}`)?

unguiculus: For secrets, the individual values are passed through `tpl`.

See https://github.com/codecentric/helm-charts/pull/250/files#diff-78f4972573ad8f1af134676ca6416fb4R25

scottrigby: Ah right, yep 👍  thanks

PR #253, opened 2020-07-15
[keycloak] Add `tpl` to Keycloak DB Config

rblaine95:
Add `tpl` to `keycloak.persistence.dbName|dbHost`
The reason for this is so that you can, for example, use utils to generate the DB Name and Host

For example:
```yaml
# values.yaml
global:
  postgres:
    host: my-example-aurora.cluster-0zZBHhzWt0rI.us-east-1.rds.amazonaws.com

keycloak:
  persistence:
    dbName: keycloak_{{ .Release.Namespace }}
    dbHost: {{ .Values.global.postgres.host }}
```

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 8.2.2
+version: 8.2.3
 appVersion: 10.0.0
 description: Open Source Identity and Access Management For Modern Applications and Services
 keywords:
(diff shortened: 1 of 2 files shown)


rblaine95:
I'll fix the conflicts asap ❤️

rblaine95:
Ok, after looking into what has changed with the refactor, it looks like the way you point to a pre-existing Postgres Cluster is to set the `DB_VENDOR`, `DB_ADDR`, `DB_PORT`, `DB_DATABASE`, `DB_USER`, and `DB_PASSWORD` environment variables manually in the `extraEnv` block in `values.yaml`.

Which renders this PR no longer needed :smiley:

PR #227, opened 2020-05-15
Default value for keycloak.persistence.existingSecretUsernameKey

f84anton:
https://github.com/codecentric/helm-charts/blob/master/charts/keycloak/templates/_helpers.tpl#L138-L146

PR #254, opened 2020-07-17
Fix missmatch between README and _helper.tpl

### LINKED ISSUE #191: keycloak.persistence.existingSecretUsernameKey seems to be set to `username` whatever happens
I am trying to install the keycloak chart with the following configuration:
```
- name: keycloak
  helm:
    chart:
      name: keycloak
      repo: https://codecentric.github.io/helm-charts
    values:
      clusterDomain: ${CLUSTER_DOMAIN}
      keycloak:
        image:
          tag: 9.0.0
        username: admin
        password: admin
        persistence:
          deployPostgres: false
          existingSecret: postgres-postgresql
          existingSecretPasswordKey: postgresql-password
          existingSecretUsernameKey: ""
          dbVendor: postgres
          dbName: keycloak
          dbHost: postgres-pgpool
          dbPort: 5432
          dbUser: postgres
```
I want to use a postgresql instance that already exists on my cluster and which defines a secret, in which I only have the database password, but not the database username (that's how [bitnami's chart](https://hub.helm.sh/charts/bitnami/postgresql-ha) works). In your documentation, you state
```
keycloak.persistence.existingSecretUsernameKey: The key for the database username in the existing secret (if keycloak.persistence.deployPostgres=false and keycloak.persistence.existingSecret != ""). Will default to the value of .keycloak.persistence.dbUser if left unset.
```
Therefore, in my config, I set
```
existingSecretUsernameKey: ""
```
making it "left unset". Therefore, I expect the `dbUser` value to be used as the username. Upon installation, I get the following error message:
```
Error: Couldn't find key username in Secret dev/postgres-postgresql
```
What do I have to do to let this helm installation use the `dbUser` I want with my existing password?

sfozz:
in the README the param keycloak.persistence.existingSecretUsernameKey is
supposed to default to the value of .keycloak.persistence.dbUser if it is
left unset.

Fixes #191

Signed-off-by: Steve Foster <stephen.foster@taitradio.com>

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 8.2.2
+version: 8.2.3
 appVersion: 10.0.0
 description: Open Source Identity and Access Management For Modern Applications and Services
 keywords:
--- a/charts/keycloak/templates/_helpers.tpl
+++ b/charts/keycloak/templates/_helpers.tpl
@@ -178,11 +178,16 @@ Create environment variables for database configuration.
   value: {{ .Values.keycloak.persistence.dbPort | quote }}
 - name: DB_DATABASE
   value: {{ .Values.keycloak.persistence.dbName | quote }}
+{{- if .Values.keycloak.persistence.existingSecretUsernameKey }}
 - name: DB_USER
   valueFrom:
     secretKeyRef:
       name: {{ include "keycloak.dbSecretName" . }}
       key: {{ include "keycloak.dbUserKey" . | quote }}
+{{- else }}
+- name: DB_USER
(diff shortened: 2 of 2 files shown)


unguiculus:
Thanks for the PR. The implementation is correct. The readme would have to be adjusted.

See:
* https://github.com/codecentric/helm-charts/blob/47332bee66dd7390d9f79eb5c52093432a034bc1/charts/keycloak/templates/_helpers.tpl#L140-L145
* https://github.com/codecentric/helm-charts/blob/47332bee66dd7390d9f79eb5c52093432a034bc1/charts/keycloak/templates/secret-db.yaml#L10-L11

sfozz:
OK so there is #227 that addresses that but it has not been signed off...

PR #258, opened 2020-07-20
[keycloak] ingress documentation

andrew-landsverk-win:
Adding some more comments to extraEnv
to help those that have ingress enabled

Signed-off-by: Andrew Landsverk <andrew.landsverk@wintechnology.com>


andrew-landsverk-win:
Looks like lint failed due to chart version needing a bump.

Also this PR was inspired by StackOverflow user Igor: https://stackoverflow.com/a/58306107/1188810

srueg:
In my tests, setting `PROXY_ADDRESS_FORWARDING=true` was enough to get ingress working correctly...

charts/keycloak/values.yaml
```diff
@@ -108,6 +108,7 @@ keycloak:
     -XX:MaxRAMPercentage=50.0
 
   ## Allows the specification of additional environment variables for Keycloak
+  ## Enable KEYCLOAK_HTTP_PORT, KEYCLOAK_HTTPS_PORT and KEYCLOAK_HOSTNAME if you are using ingress: true
```
srueg, line 111:
```suggestion
  ## Enable PROXY_ADDRESS_FORWARDING if you are using ingress: true
```

andrew-landsverk-win: I will test again on my dev cluster to see, thanks for the input!

andrew-landsverk-win: So, it deploys right, and you can access the welcome page at mydomain.com/auth/ but the links to the realm management don't work because they have ":8443" at the end of them (even removing the ports and trying to navigate directly doesn't work), so you actually need all 4 env variables set to get ingress to function properly.

- `KEYCLOAK_HTTP_PORT`
- `KEYCLOAK_HTTPS_PORT`
- `KEYCLOAK_HOSTNAME`
- `PROXY_ADDRESS_FORWARDING`

srueg:
May I ask which ingress controller you're using? For me it works with the [nginx-ingress](https://github.com/kubernetes/ingress-nginx/).

According to the [docs](https://www.keycloak.org/docs/latest/server_installation/#_setting-up-a-load-balancer-or-proxy) the LB needs to set the `X-Forwarded-For` and `X-Forwarded-Proto` headers but I think it also needs the `X-Forwarded-Port` header to work correctly in this case.

If I access `/auth/realms/master/.well-known/openid-configuration` I get the right URLs and ports.

andrew-landsverk-win:
You were correct :) I was using kong as the ingress controller. They just pushed an updated chart which fixes the port mapping issues. I will update my PR to reflect the suggestion you mentioned. Thank you!

PR #259, opened 2020-07-21
[keycloak] Add template capabilities on ingress host names

jBouyoud:
Add template capabilities on ingress host names.

Eg.:  In order to use `Release.Name` in  host name


jBouyoud:
@thomasdarimont @unguiculus Anything that can I do to help merge this PR ?

PR #262, opened 2020-07-24
[keycloak] Add network policies support

jBouyoud:
Allow ingress network policies for keycloak and postgres

### DIFF
--- a/charts/keycloak/README.md
+++ b/charts/keycloak/README.md
@@ -108,6 +108,8 @@ The following table lists the configurable parameters of the Keycloak chart and 
 | `ingress.tls` | TLS configuration | see below |
 | `ingress.tls[0].hosts` | List of TLS hosts | `[keycloak.example.com]` |
 | `ingress.tls[0].secretName` | Name of the TLS secret | `""` |
+| `networkPolicy.enabled` | If true, the ingress network policy is deployed | `false`
+| `networkPolicy.extraFrom` | Allows to define allowed external traffic (see Kubernetes doc for network policy `from` format) | `[]`
 | `route.enabled` | If `true`, an OpenShift Route is created | `false` |
 | `route.path` | Path for the Route | `/` |
 | `route.annotations` | Route annotations | `{}` |
--- a/charts/keycloak/ci/postgres-ha-netpol-values.yaml
+++ b/charts/keycloak/ci/postgres-ha-netpol-values.yaml
@@ -1,0 +1,90 @@
+replicas: 2
+
+podLabels:
+  test-label: test-label-value
+
+podAnnotations:
+  test-annotation: test-annotation-value-{{ .Release.Name }}
+  test-int-annotation: "12345"
+
+startupScripts:
+  hello.sh: |
+    #!/bin/sh
+
+    echo '********************************************************************************'
+    echo '*                                                                              *'
+    echo '*                        Hello from my startup script!                         *'
+    echo '*                                                                              *'
+    echo '********************************************************************************'
+
+lifecycleHooks: |
+  postStart:
+    exec:
+      command:
+        - /bin/sh
+        - -c
+        - echo 'Hello from lifecycle hook!'
+
+extraEnv: |
+  - name: JGROUPS_DISCOVERY_PROTOCOL
+    value: dns.DNS_PING
+  - name: JGROUPS_DISCOVERY_PROPERTIES
+    value: 'dns_query={{ include "keycloak.serviceDnsName" . }}'
+  - name: CACHE_OWNERS_COUNT
+    value: "2"
+  - name: CACHE_OWNERS_AUTH_SESSIONS_COUNT
+    value: "2"
+  - name: KEYCLOAK_USER_FILE
+    value: /secrets/admin-creds/user
+  - name: KEYCLOAK_PASSWORD_FILE
+    value: /secrets/admin-creds/password
+  - name: KEYCLOAK_STATISTICS
+    value: all
+  - name: JAVA_OPTS
+    value: >-
+      -XX:+UseContainerSupport
+      -XX:MaxRAMPercentage=50.0
+      -Djava.net.preferIPv4Stack=true
+      -Djboss.modules.system.pkgs=$JBOSS_MODULES_SYSTEM_PKGS
+      -Djava.awt.headless=true
+
+secrets:
+  admin-creds:
+    stringData:
+      user: admin
+      password: secret
+
+networkPolicy:
+  enabled: true
+  labels:
+    my-custom-label: "true"
+
+  extraFrom:
+    - namespaceSelector:
+        matchLabels:
+          ingress: "true"
+
+extraPorts:
+  - name: http-admin
+    containerPort: 8444
+    protocol: TCP
+
+extraVolumeMounts: |
+  - name: admin-creds
+    mountPath: /secrets/admin-creds
+    readOnly: true
+
+extraVolumes: |
+  - name: admin-creds
+    secret:
+      secretName: '{{ include "keycloak.fullname" . }}-admin-creds'
+
+postgresql:
+  enabled: true
+  persistence:
+    enabled: true
+  networkPolicy:
+    enabled: true
+
+test:
+  enabled: true
--- a/charts/keycloak/templates/networkpolicy.yaml
+++ b/charts/keycloak/templates/networkpolicy.yaml
@@ -1,0 +1,46 @@
+{{- if .Values.networkPolicy.enabled }}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: {{ include "keycloak.fullname" . | quote }}
+  labels:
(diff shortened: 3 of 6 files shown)


jBouyoud:
@thomasdarimont @unguiculus Anything that can I do to help merge this PR ?

charts/keycloak/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,46 @@
...
+      {{- include "keycloak.selectorLabels" . | nindent 6 }}
+  ingress:
+    {{- if .Values.networkpolicy.from }}
+    - from:
+        {{- toYaml .Values.networkpolicy.from | nindent 8 }}
+      ports:
+        - protocol: TCP
+          port: {{ .Values.service.httpPort }}
+        - protocol: TCP
+          port: {{ .Values.service.httpsPort }}
```
unguiculus, line 25: Use named ports. Also add the management port.

charts/keycloak/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,46 @@
...
+        - protocol: TCP
+          port: {{ .Values.service.httpManagementPort }}
+        {{ with .Values.extraPorts }}
+        - protocol: {{ default "TCP" .protocol }}
+          port: {{ .containerPort }}
+        {{- end }}
```
unguiculus, line 45: See above

charts/keycloak/templates/statefulset.yaml
```diff
@@ -32,6 +32,11 @@ spec:
         {{- end }}
       labels:
         {{- include "keycloak.selectorLabels" . | nindent 8 }}
+        {{- if and .Values.postgresql.enabled (and .Values.postgresql.networkPolicy .Values.postgresql.networkPolicy.enabled) }}
+        # Mimic Postgres context in order to reuse postgresql.fullname template and not hard code this value
+        {{- $chartValue := dict "Chart" (dict "Name" "postgresql") "Values" (mustDeepCopy .Values.postgresql) "Release" .Release }}
+        "{{ template "postgresql.fullname" $chartValue }}-client": "true"
+        {{- end }}
```
unguiculus, line 39:
```suggestion
        {{- if and .Values.postgresql.enabled (and .Values.postgresql.networkPolicy .Values.postgresql.networkPolicy.enabled) }}
        {{ include "keycloak.postgresql.fullname" . }}-client: "true"
        {{- end }}
```

charts/keycloak/values.yaml
```diff
@@ -261,6 +261,15 @@ ingress:
...
       secretName: keycloak-tls
 
+## Network policies configuration
+networkpolicy:
+  # If true, the Network policies are deployed
+  enabled: false
+
+  # Define all other external allowed source
+  # See https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#networkpolicypeer-v1-networking-k8s-io
+  # from :{}
```
unguiculus, line 271:
```suggestion
  from: {}
```

charts/keycloak/README.md
```diff
@@ -108,6 +108,8 @@ The following table lists the configurable parameters of the Keycloak chart and
 | `ingress.tls` | TLS configuration | see below |
 | `ingress.tls[0].hosts` | List of TLS hosts | `[keycloak.example.com]` |
 | `ingress.tls[0].secretName` | Name of the TLS secret | `keycloak-tls` |
+| `networkpolicy.enabled` |If true, the ingress network policy is deployed | `false`
+| `networkpolicy.from` |Allow to define allowed external traffic (see Kubernetes doc for network policy `from` format) | ``
```
unguiculus, line 112:
```suggestion
| `networkpolicy.from` | Allows to define allowed external traffic (see Kubernetes doc for network policy `from` format) | ``
```

unguiculus: Maybe call it `networkpolicy.extraFrom` because it might be confusing as it's optional.

unguiculus (changes requested):

charts/keycloak/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,46 @@
...
+  ingress:
+    {{- if .Values.networkpolicy.from }}
+    - from:
+        {{- toYaml .Values.networkpolicy.from | nindent 8 }}
+      ports:
+        - protocol: TCP
+          port: {{ .Values.service.httpPort }}
+        - protocol: TCP
+          port: {{ .Values.service.httpsPort }}
+        {{ with .Values.extraPorts }}
```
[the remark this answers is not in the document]
jBouyoud: My bad, typo this is a `range` not `with`

charts/keycloak/templates/networkpolicy.yaml
jBouyoud: Named port on network policies is not support by all k8s networking layer. 
For exemple, my current cluster doesn't support it (weave-net, old version :sad: ), if someone define a netpol with a named port this break the entire cluster networking.

Since this is a chart (aka managed by value files), name or configured port is quite the same, no ?
Or maybe I can add an option to manage this (useNamedPort: boolean). 
Let me know what is your preference ? 

For managed port, this is intentional. 
I was thinking that port should only be used by Keycloak pods. That's probabily wrong. 
Should I add an option for each port to let user define which port is allowed on this `from` user defined policies ? 

`managementPort` is already present on keycloak to keycloak pods communication policy (below)

charts/keycloak/values.yaml
jBouyoud: Empty and not defined `from` doesn't mean the same thing in network policies. 
That why i do not se a default value for this key and I add a `{{- if .Values.networkpolicy.from }}` in the template.

charts/keycloak/templates/networkpolicy.yaml
unguiculus: I didn't try it but it is documented that named ports work:

```
$ k explain networkpolicy.spec.ingress.ports
KIND:     NetworkPolicy
VERSION:  networking.k8s.io/v1

RESOURCE: ports <[]Object>

DESCRIPTION:
     List of ports which should be made accessible on the pods selected for this
     rule. Each item in this list is combined using a logical OR. If this field
     is empty or missing, this rule matches all ports (traffic not restricted by
     port). If this field is present and contains at least one item, then this
     rule allows traffic only if the traffic matches at least one port in the
     list.

     NetworkPolicyPort describes a port to allow traffic on

FIELDS:
   port	<string>
     The port on the given protocol. This can either be a numerical or named
     port on a pod. If this field is not provided, this matches all port names
     and numbers.

   protocol	<string>
     The protocol (TCP, UDP, or SCTP) which traffic must match. If not
     specified, this field defaults to TCP.
```

charts/keycloak/values.yaml
unguiculus: Actuall, it should be a list. I was wrong here. So, the default should be `[]`. It is best practice to always specify a default. this should work with your existing code. Just keep the `if`. Actually, a `with` statement would be preferable here.

charts/keycloak/templates/networkpolicy.yaml
jBouyoud: Yes, totally agreed but their is some issues on implementations : https://github.com/weaveworks/weave/issues/3785. 
If I put named port in this PR a lot of users cannot be able to use this networkpolicies template (like me ;-) )

charts/keycloak/ci/postgres-ha-netpol-values.yaml
```diff
@@ -0,0 +1,90 @@
...
+      -Djboss.modules.system.pkgs=$JBOSS_MODULES_SYSTEM_PKGS
+      -Djava.awt.headless=true
+
+secrets:
+  admin-creds:
+    stringData:
+      user: admin
+      password: secret
+
+networkPolicy:
```
unguiculus, line 57: This was actuall incorrect, but now is, so the test did not have any effect.

charts/keycloak/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,46 @@
+{{- if .Values.networkpolicy.enabled }}
+apiVersion: "networking.k8s.io/v1"
+kind: "NetworkPolicy"
+metadata:
+  name: "{{ include "keycloak.fullname" . }}"
+  labels:
+    {{- include "keycloak.labels" . | nindent 4 }}
+    {{- range $key, $value := .Values.networkpolicy.labels }}
```
unguiculus, line 8:
```suggestion
    {{- range $key, $value := .Values.networkPolicy.labels }}
```

unguiculus (changes requested):
Almost good to go 😉

charts/keycloak/ci/postgres-ha-netpol-values.yaml
jBouyoud: Well spotted, I'll check that on future run

jBouyoud:
Looks good for me

unguiculus (approved):

Issue #271, opened 2020-07-30
cannot enter admin console after new installation of keycloak chart v9.0.1


YoshieraHuang:
I upgraded the chart to v9.0.1. I used chrome to enter the keycloak pages. However i cannot enter the admin console. Here is the error at browser console:
```
Mixed Content: The page at 'https://auth.pims.private/auth/admin/master/console/' was loaded over HTTPS, but requested an insecure script 'http://auth.pims.private/auth/js/keycloak.js?version=bouyi'. This request has been blocked; the content must be served over HTTPS.
```
![Screen Capture_select-area_20200730213123](https://user-images.githubusercontent.com/38752027/88928961-06e37800-d2ac-11ea-9ad8-9b51b0c7218c.png)

It seems that the secure page is forbidden to access insecure contents in chrome. Is this a bug of this charts or the keycloak application?

Here is my custom values for chart:
```
extraEnv: |
  - name: KEYCLOAK_USER
    value: pims_private
  - name: KEYCLOAK_PASSWORD
    value: XXXXXX
serviceAccount:
  create: true
ingress:
  enabled: true
  annotations:
    # cert-manager to sign certificate
    cert-manager.io/cluster-issuer: ca-cluster-issuer
    kubernetes.io/ingress.class: "nginx"
    nginx.ingress.kubernetes.io/enable-cors: "true"
    nginx.ingress.kubernetes.io/cors-allow-origin: "*.pims.private"
  rules:
  - host: auth.pims.private
    paths: [/]
  tls:
  - hosts:
    - auth.pims.private
    secretName: keycloak-cert

postgresql:
  postgresqlPassword: XXXXXXX
  persistence:
    storageClass: managed-nfs-storage
    enabled: true
```

ifalex:
Encountering same issue.

gallagth:
Same issue here

CalamarBicefalo:
Looks like they removed the proxy forwarding option:
https://github.com/codecentric/helm-charts/commit/8b8044aa4458ca4a71d5420a503bedb2444ed605#diff-f287956fb12d7d7c321c313e858d887aL102

Maybe we gotta set the env var instead? Just like they say we have to do with username/pass.

CalamarBicefalo:
Try adding this @ifalex @gallagth @YoshieraHuang 

```yaml
extraEnv: |
    - name: PROXY_ADDRESS_FORWARDING
      value: true
```

[...]

PR #275, opened 2020-08-03
[keycloak] fix object type of extraEnv in schema

jonkerj:
Keycloaks `values.yaml.schema` incorrectly forced `extraEnv` to be a string, which is kind of silly, as it needs to be an array containing `.name` and `.value` items.

Technically fixes codecentric/helm-charts#271

### DIFF
--- a/charts/keycloak/values.schema.json
+++ b/charts/keycloak/values.schema.json
@@ -56,7 +56,14 @@
       "type": "string"
     },
     "extraEnv": {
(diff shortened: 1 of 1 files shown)


unguiculus:
In fact, it is not silly but on purpose. Please read the docs:
https://github.com/codecentric/helm-charts/tree/master/charts/keycloak#usage-of-the-tpl-function

PR #305, opened 2020-09-15
Update README.md

genbutsu:
since many Keycloak release this settings seems necessary to support https access via ingress


genbutsu:
wrong place for the commit :-(

Issue #303, opened 2020-09-10
Keycloak chart 9.0.7 not working under https


codependent:
I've deployed the latest version of the Keycloak chart:

```
helm install keycloak codecentric/keycloak --set ingress.enabled=true --set ingress.rules[0].host=myhost.io --set ingress.rules[0].paths={/} --set ingress.tls[0].hosts={myhost.io}  --namespace=lab
```

When everything is up and running I access https://myhost.io. After setting the admin user and password I try to access the admin console: https://myhost.io/auth/admin/master/console/

However the browser never loads the page since the load of the keycloak.js file is interrupted. It is trying to get it from http:// and given the browser URL is https it blocks that content.

Source of the https://myhost.io/auth/admin/master/console/ page, showing that the keycloak.js resource is gotten using http.

```
    <script src="/auth/resources/9dhzl/common/keycloak/lib/ui-ace/min/ace.js"></script>
    <script src="/auth/resources/9dhzl/common/keycloak/lib/ui-ace/ui-ace.min.js"></script>

    <script src="http://myhost.io/auth/js/keycloak.js?version=9dhzl" type="text/javascript"></script>
```
Is there anything missing in the chart to make it work with https o is it a Keycloak problem?

unguiculus:
You might have to add annotations depending on the ingress controller you use. Also, you did not specify a TLS secret.

codependent:
Regarding the secret, I have configured the nginx ingress controller to use a default secret, no problem there. Regarding the annotations I’ll have a look at them. Oddly enough it works without anything else up to chart version 8.3.0.

unguiculus:
There was an issue with the TLS cert (#281, #288) always being added which has been fixed.

codependent:
Thanks for letting me know. Any idea when the fix will be released?

unguiculus:
It has already been released.

codependent:
I thought this problem was solved but after trying again with both chart versions 9.0.8 and 9.1.0 it still happens:

A clean installation with 9.1.0...
```
helm install keycloak codecentric/keycloak --set ingress.enabled=true --set ingress.rules[0].host=lab.x.io --set ingress.rules[0].paths={/} --set ingress.tls[0].hosts={lab.x.io} --namespace=lab
```
...creates an ingress like this (notice I don't specify a tls secret since the nginx controller has a default one configured):
```
apiVersion: v1
items:
- apiVersion: extensions/v1beta1
  kind: Ingress
  metadata:
[...]

PR #306, opened 2020-09-15
Update README.md

genbutsu:
At least from version 9 the connection will be broken as in issue#303; I've got the same problem with the last version of Keycloak and I've found and tested this solution

https://github.com/codecentric/helm-charts/issues/303


unguiculus:
Superseded by #322. Thanks.

PR #318, opened 2020-10-08
Update README.md

hronix:
According to variables in templates, **port** should be instead of **node**


unguiculus:
Thanks. Please fix DCO and bump the chart version so this can be merged.

PR #320, opened 2020-10-09
[keycloak] Do not set enableServiceLinks when false

BrentVLaere:
The specification of `enableServiceLinks` gives an error on older Kubernetes versions.
`enableServiceLinks` is now only specified when its value is false


unguiculus:
`enableServiceLinks` has been available since Kubernetes 1.13. 1.16 isn't even officially supported anymore. I don't intend to extend support beyond what's specified here: https://kubernetes.io/docs/setup/release/version-skew-policy/.

PR #323, opened 2020-10-14
[keycloak] Update ingress API version

shane-davidson:
Update ingress version to work with kubernetes 1.19

uses a helper function to select v1 or v1beta1

### DIFF
--- a/charts/keycloak/templates/_helpers.tpl
+++ b/charts/keycloak/templates/_helpers.tpl
@@ -87,3 +87,15 @@ Create the namespace for the ServiceMonitor deployment
 {{ .Release.Namespace }}
 {{- end }}
 {{- end }}
+
+
+{{/*
+Return the appropriate apiVersion for ingress.
+*/}}
+{{- define "ingress.apiVersion" -}}
+{{- if semverCompare ">=1.19.0" .Capabilities.KubeVersion.GitVersion -}}
+{{- print "networking.k8s.io/v1" -}}
+{{- else -}}
+{{- print "networking.k8s.io/v1beta1" -}}
+{{- end -}}
+{{- end -}}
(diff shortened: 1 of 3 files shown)


charts/keycloak/templates/ingress.yaml
```diff
@@ -1,6 +1,6 @@
 {{- $ingress := .Values.ingress -}}
 {{- if $ingress.enabled -}}
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: {{ template "ingress.apiVersion" . }}
```
unguiculus, line 3:
```suggestion
apiVersion: {{ include "keycloak.ingressAPIVersion" . }}
```

charts/keycloak/templates/_helpers.tpl
```diff
@@ -87,3 +87,15 @@ Create the namespace for the ServiceMonitor deployment
 {{ .Release.Namespace }}
 {{- end }}
 {{- end }}
+
+
+{{/*
+Return the appropriate apiVersion for ingress.
+*/}}
+{{- define "ingress.apiVersion" -}}
```
unguiculus, line 95:
```suggestion
{{- define "keycloak.ingressAPIVersion" -}}
```

charts/keycloak/templates/_helpers.tpl
```diff
@@ -87,3 +87,15 @@ Create the namespace for the ServiceMonitor deployment
...
+{{- if semverCompare ">=1.19.0" .Capabilities.KubeVersion.GitVersion -}}
```
unguiculus, line 96: Use this instead:
```suggestion
{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" -}}
```

unguiculus (changes requested):
Thanks. Please bump the chart version and fix DCO.

PR #328, opened 2020-10-19
[keycloak] Add the global.imageRegistry option

yardenshoham:
This change allows one to specify the image registry which will be used for each image individually, as well as a global image registry to override all image registries.

I used Bitnami's common chart to keep things standard and generic.

This change is extremely beneficial to those who use an airgapped environment with a private image registry.

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 9.4.0
+version: 10.0.0
 appVersion: 11.0.2
 description: Open Source Identity and Access Management For Modern Applications and Services
 keywords:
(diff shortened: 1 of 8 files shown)


yardenshoham:
@thomasdarimont @unguiculus please review and merge.

unguiculus (changes requested):
I'm open to adding a separate registry configuration option, but I certainly won't accept including the common chart dependency. Also, this should be done in a backwards-compatible way so we don't need to bump the major version.

yardenshoham:
@unguiculus 
Currently the `image.repository` includes the registry. There is no way to introduce `image.registry` without changing the meaning of `image.repository`. So there is no way to make this change be backwards-compatible.

yardenshoham:
@unguiculus What do you think?

unguiculus:
Please remove the common chart dependency and I'll have another look.

PR #340, opened 2020-11-06
add condition for ingress apiVersion

cgroschupp:
k8s < 1.14 only accept api version `extensions/v1beta1` for ingress:
https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 9.5.0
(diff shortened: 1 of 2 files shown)


unguiculus:
Currently, the offically supported Kubernetes versions are 1.17+ (https://kubernetes.io/docs/setup/release/version-skew-policy/). This chart will not offer support beyond that. You should really upgrade.

cgroschupp:
@unguiculus Thank you for your hint, but this change is very small and very common in many charts.

PR #344, opened 2020-11-16
[mailhog] Introduce `.Release.Namespace` in objects meta

bkondek:
Description of the change

This PR adds .Release.Namespace to objects meta for Mailhog chart.

Benefits

Ability to use the helm chart when having declarative definition of all cluster objects rendered using helm template, for example in tools like Spinnaker.

Possible drawbacks

I don't see any, it will work as previously for helm install.

Signed-off-by: Bartosz Kondek <b.kondek@piwik.pro>

### DIFF
--- a/charts/mailhog/templates/ingress.yaml
+++ b/charts/mailhog/templates/ingress.yaml
@@ -9,8 +9,8 @@ metadata:
     helm.sh/chart: {{ include "mailhog.chart" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
-  {{- with .Values.ingress.annotations }}
   namespace: {{ .Release.Namespace }}
+  {{- with .Values.ingress.annotations }}
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
--- a/charts/mailhog/templates/ingress.yaml
+++ b/charts/mailhog/templates/ingress.yaml
@@ -10,6 +10,7 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
   {{- with .Values.ingress.annotations }}
+  namespace: {{ .Release.Namespace }}
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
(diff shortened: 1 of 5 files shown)


mrsegen:
I may be missing something due to unfamiliarity with Spinnaker, but: Isn't the namespace for a release's contents automatically set by Helm when the release is installed?

A few spots on [this FAQ page](https://helm.sh/docs/faq/#release-names-are-now-scoped-to-the-namespace) speak to it.

Though, I agree that I can't see any drawbacks.

mrsegen (approved):

unguiculus:
IMO, the namespace should not be part of the manifests. You specify it when you install the chart or the rendered templates. It is very uncommon the add the namespace to the manifests. If Spinnaker does not allow installing into a namespace without it being added to the manifest, then that's something that should be fixed on the Spinnaker side.

glothriel:
Hi @unguiculus @mrsegen 

It's quite important for us, could you spare a minute and read this thread? https://github.com/bitnami/charts/issues/2006, it provides more context.

mrsegen:
I appreciate the context, the effort to fully declare a cluster, and the included tip about Sealed Secrete!

I didn't dive into the discussions that also linked to that issue, but IIUC helm2 may be _part_ of the sticking point. If that's the case, perhaps this change should only be temporary.

Though, given that:
- that my strongest case against this change is "It's more elegant to not specify more than necessary,"
- including meta.namespace apparently **is** necessary for your use case, and
- Kubernetes is evolving and collaborate, calling for participation and flexibility of all

... my (unweighted 😆) approval stands.

charts/mailhog/templates/ingress.yaml
```diff
@@ -10,6 +10,7 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
   {{- with .Values.ingress.annotations }}
+  namespace: {{ .Release.Namespace }}
```
unguiculus, line 13:
```suggestion
  namespace: {{ .Release.Namespace }}
  {{- with .Values.ingress.annotations }}
```

unguiculus (changes requested):
Alright, I'll consider merging it.

unguiculus (approved):

glothriel:
Thanks guys!

PR #348, opened 2020-11-20
[mailhog] Add ability to set Web UI path

bkondek:
Added ability to setup custom Web UI path, which will allow
to use application in context another than "/" in ingress

Signed-off-by: Bartosz Kondek <b.kondek@piwik.pro>

### DIFF
--- a/charts/mailhog/Chart.yaml
+++ b/charts/mailhog/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 description: An e-mail testing tool for developers
 name: mailhog
 appVersion: v1.0.1
-version: 3.3.0
+version: 3.4.0
 keywords:
   - mailhog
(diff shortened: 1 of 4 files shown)


bkondek:
@unguiculus could you please look that this PR?
I really need that variable to use this chart with ingress

glothriel:
@mrsegen @unguiculus Sorry guys for pinging again, but it's quite important for as, we would really like to merge this.

unguiculus:
Sorry for the delayed response. The chart already supports adding additional environment variables.

glothriel:
Indeed, there is `extraEnv` variable. Imo we can close this, @bkondek ?

bkondek:
Ok, thanks, i've missed this.
In this case - closing this PR

PR #350, opened 2020-11-23
[keycloak] Allow setting loadBalancerSourceRanges on the svc

marcin-je:
When creating Service with type of LoadBalancer it results with Security
Group (or equivalent) open to 0.0.0.0/0. By allowing to define
`loadBalancerSourceRanges` one can limit that to a list of allowed
CIDRs. This is a must when a cloud policy of an organisation doesn't
allow opening security groups to 0.0.0.0/0.

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 9.6.2
+version: 9.7.0
 appVersion: 11.0.2
 description: Open Source Identity and Access Management For Modern Applications and Services
 keywords:
--- a/charts/keycloak/templates/service-http.yaml
+++ b/charts/keycloak/templates/service-http.yaml
@@ -19,6 +19,10 @@ spec:
   {{- if and (eq "LoadBalancer" .Values.service.type) .Values.service.loadBalancerIP }}
   loadBalancerIP: {{ .Values.service.loadBalancerIP }}
   {{- end }}
+  {{- if and (eq "LoadBalancer" .Values.service.type) .Values.service.loadBalancerSourceRanges }}
+  loadBalancerSourceRanges:
+    {{- toYaml .Values.service.loadBalancerSourceRanges | nindent 4 }}
+  {{- end }}
   ports:
     - name: http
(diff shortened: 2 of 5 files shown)


charts/keycloak/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 9.6.1
+version: 9.6.2
```
unguiculus, line 3: Please bump the minor version.

charts/keycloak/templates/service-http.yaml
```diff
@@ -19,6 +19,10 @@ spec:
   {{- if and (eq "LoadBalancer" .Values.service.type) .Values.service.loadBalancerIP }}
   loadBalancerIP: {{ .Values.service.loadBalancerIP }}
   {{- end }}
+  {{- if and (eq "LoadBalancer" .Values.service.type) .Values.service.loadBalancerSourceRanges }}
+  loadBalancerSourceRanges:
+  {{ toYaml .Values.service.loadBalancerSourceRanges | indent 2 }}
+  {{- end }}
```
unguiculus, line 25:
```suggestion
  {{- if and (eq "LoadBalancer" .Values.service.type) .Values.service.loadBalancerSourceRanges }}
  loadBalancerSourceRanges:
    {{- toYaml .Values.service.loadBalancerSourceRanges | nindent 4 }}
  {{- end }}
```

unguiculus (changes requested):
Please update the readme as well.

marcin-je:
@unguiculus thanks for review, all done ✅

charts/keycloak/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 9.6.2
+version: 9.6.3
```
unguiculus, line 3:
```suggestion
version: 9.7.0
```

unguiculus (changes requested):

marcin-je:
@unguiculus thanks, version bumped as requested

unguiculus (approved):

Issue #349, opened 2020-11-20
unable to recognize "": no matches for kind "Ingress" in version "networking.k8s.io/v1"


barnybadzoo:
Hi after the last commit i receive this error message if i want to install per helm:

unable to recognize "": no matches for kind "Ingress" in version "networking.k8s.io/v1"

I try to install on k3OS v0.19.2-dev.4 .
My Versions:

Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"**v1.18.6+k3s1**", GitCommit:"6f56fa1d68a5a48b8b6fdefa8eb7ead2015a4b3a", GitTreeState:"clean", BuildDate:"2020-07-16T20:46:15Z", GoVersion:"go1.13.11", Compiler:"gc", Platform:"linux/amd64"}

Till yesterday everything worked fine..

barnybadzoo:
Workaround: Install an older Version with the Versionflag:

> sudo helm install keycloak codecentric/keycloak --version "9.5.0" --namespace keycloak --values ./keycloak-values.yaml

jonkerj:
I am facing the same issue. The code in `_helpers.tpl` assumes that if `networking.k8s.io/v1` is present, `Ingress` lives there. In my cluster (1.18.10) this is not the case: `Ingress` is still `v1beta1`. I am not a helm templating wizard (otherwise I'd PR it), but it seems to me that it could be better to check the Kubernetes version instead. 

This change was introduced in #334.

Flowkap:
@jonkerj is correct. The condition is false. In 1.18 the api version exists but INgress is not yet in it. So this helm version is broken for K8s 1.18.x

seboudry:
Hi! Also have this issue with K8S 1.17 BTW

0xErnie:
As far as I understood the cause of the problem is, that only the existence of the API is checked, but not the existence of the class Ingress, so the following might solve the problem:
﻿`.Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress"`
istead of
`.Capabilities.APIVersions.Has "networking.k8s.io/v1"`

jildertmiedema:
Solved with https://github.com/codecentric/helm-charts/pull/357

unguiculus:
Thanks @jildertmiedema

Issue #347, opened 2020-11-20
[Keycloak] Invalid ingress specs


tbnguyen1407:
Hello latest helm chart (9.6.0+) includes invalid spec for networking.k8s.io/v1 ingress. This fails when deploying on k8s 1.19+

**Error:**

```
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(Ingress.spec.rules[0].http.paths[0].backend): unknown field "serviceName" in io.k8s.api.networking.v1.IngressBackend, ValidationError(Ingress.spec.rules[0].http.paths[0].backend): unknown field "servicePort" in io.k8s.api.networking.v1.IngressBackend]
```
**Cause:**

https://github.com/codecentric/helm-charts/blob/8984699ec22d1de39d31919d774bb65c957be8ef/charts/keycloak/templates/ingress.yaml#L36-L41

**Workaround:**

Downgrade to 9.5.0

**Fix:**

for v1 it should be
```
- path: {{ . }}
  pathType: PathPrefix
  backend:
    service:
      name: {{ include "keycloak.fullname" $ }}-http
      port:
        name: {{ $ingress.servicePort }}
```

**Notes:**

You may need to have separate blocks to handle v1 and v1beta1 spec for fallback. Check for exact Ingress resource instead of just networking.k8s.io/v1 since this is also available in older k8s versions

```
{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
apiVersion: networking.k8s.io/v1
{{- else }}
apiVersion: networking.k8s.io/v1beta1
{{- end }}
```

barnybadzoo:
#349 same issue

SeWieland:
Seems like using my K8s version v1.18.8, the charts render the apiVersion: networking.k8s.io/v1 as well.
Thus, the chart might be broken for versions <19 in regards to the ingress api version.

BaloMueller:
Latest version (9.6.1) also doesn't work for me.
kubectl version: v1.19.3
server version: v1.17.13

Using version 9.5.0 fixed it for me though.

Flowkap:
/duplicate https://github.com/codecentric/helm-charts/issues/349

PR #352, opened 2020-11-23
[keycloak] Fix ingress

RytisLT:
Currently, the helm chart fails when ingress is enabled. This pull request adapts ingress.yaml to ingress v1 syntax. Might be a good idea to add ingress deployment to automated testing.

### DIFF
--- a/charts/keycloak/templates/ingress.yaml
+++ b/charts/keycloak/templates/ingress.yaml
@@ -35,9 +35,19 @@ spec:
         paths:
           {{- range .paths }}
           - path: {{ . }}
+            {{- if $.Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
+            pathType: Prefix
             backend:
-              serviceName: {{ include "keycloak.fullname" $ }}-http
-              servicePort: {{ $ingress.servicePort }}
+              service:
+                name: {{ include "keycloak.fullname" $ }}-http
+                port:
+                  name: {{ $ingress.servicePort }}
+            {{- else }}
+            backend:
+              service:
+                serviceName: {{ include "keycloak.fullname" $ }}-http
+                servicePort: {{ $ingress.servicePort }}
+            {{- end }}
           {{- end }}
     {{- end }}
 {{- end -}}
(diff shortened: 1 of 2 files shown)


jkroepke:
Hi,

i'm running Kubernetes 1.19, looks like pathType is required, too:

```yaml
# Source: keycloak/templates/patched.yaml
kind: Ingress
metadata:
  annotations:
    kubernetes.io/tls-acme: "true"
    nginx.ingress.kubernetes.io/proxy-buffer-size: 128k
  labels:
    app.kubernetes.io/instance: keycloak
    app.kubernetes.io/name: keycloak
    app.kubernetes.io/version: 11.0.2
    helm.sh/chart: keycloak-9.6.1
  name: keycloak
spec:
  rules:
  - host: example.keycloak.org
    http:
      paths:
      - backend:
          service:
            name: keycloak-http
            port:
              name: http
        path: /
  tls:
  - hosts:
    - example.keycloak.org
    secretName: keycloak-ingress-tls
```

Results into this:
```
  Error: release keycloak failed, and has been uninstalled due to atomic being set: Ingress.extensions "keycloak" is invalid: spec.rules[0].http.paths[0].pathType: Required value: pathType must be specified
```

charts/keycloak/templates/ingress.yaml
```diff
@@ -36,8 +36,10 @@ spec:
           {{- range .paths }}
           - path: {{ . }}
```
jkroepke, line 37:
```suggestion
          - path: {{ . }}
            pathType: Prefix
```

unguiculus (changes requested):
This is related to this change: #334

The fix needs to consider what API version is used.

RytisLT:
@unguiculus added ingress version check. I couldn't figure out how to use `keycloak.ingressAPIVersion` inside `if` statement so I've used `.Capabilities` instead.

RytisLT:
@unguiculus Can you please take another look at the PR I've added the version check?

Flowkap:
The new versioh will break evetyrhing before K8s 1.19.x so should be v1beta1 still. Related to https://github.com/codecentric/helm-charts/issues/349 https://github.com/codecentric/helm-charts/issues/347

unguiculus (approved):

PR #353, opened 2020-11-30
[mailhog] add ability to set labels on the ingress object

tobylo:
Add possiblity to set labels on the ingress resource created. This is necessary e.g. when running multiple ingress controllers in a cluster which are using labelSelectors (as it is in our case)

Signed-off-by: Tobias Lolax <tobias.lolax@gmail.com>

### DIFF
--- a/charts/mailhog/Chart.yaml
+++ b/charts/mailhog/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 description: An e-mail testing tool for developers
 name: mailhog
 appVersion: v1.0.1
-version: 3.3.0
+version: 3.3.1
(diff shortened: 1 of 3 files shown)


charts/mailhog/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 description: An e-mail testing tool for developers
 name: mailhog
 appVersion: v1.0.1
-version: 3.3.0
+version: 3.3.1
```
unguiculus, line 5: Please bump the minor version.

unguiculus (changes requested):

PR #355, opened 2020-12-03
[keycloak] Add Affinity to HTTP Service

s-vkropotko:
Signed-off-by: Viacheslav Kropotko <vjkropotko@gmail.com>

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 9.7.0
+version: 9.8.0
 appVersion: 11.0.2
 description: Open Source Identity and Access Management For Modern Applications and Services
 keywords:
--- a/charts/keycloak/templates/service-http.yaml
+++ b/charts/keycloak/templates/service-http.yaml
@@ -25,9 +25,9 @@ spec:
   {{- end }}
   {{- if .Values.service.sessionAffinity }}
   sessionAffinity: {{ .Values.service.sessionAffinity }}
-  {{- with .Values.service.sessionAffinityConfig }}
+    {{- with .Values.service.sessionAffinityConfig }}
   sessionAffinityConfig:
-  {{- toYaml . | nindent 4 }}
+    {{- toYaml . | nindent 4 }}
   {{- end }}
   {{- end }}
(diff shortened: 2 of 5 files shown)


charts/keycloak/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 9.6.1
+version: 9.6.2
```
unguiculus, line 3: Should be a minor bump

charts/keycloak/templates/service-http.yaml
```diff
@@ -19,6 +19,12 @@ spec:
   {{- if and (eq "LoadBalancer" .Values.service.type) .Values.service.loadBalancerIP }}
   loadBalancerIP: {{ .Values.service.loadBalancerIP }}
   {{- end }}
+  {{- if .Values.service.sessionAffinity }}
+  sessionAffinity: {{ .Values.service.sessionAffinity }}
+  {{- if .Values.service.sessionAffinityConfig }}
+  sessionAffinityConfig: {{- toYaml .Values.service.sessionAffinityConfig | nindent 4 }}
+  {{- end }}
```
unguiculus, line 26:
```suggestion
  {{- with .Values.service.sessionAffinityConfig }}
  sessionAffinityConfig: 
  	{{- toYaml . | nindent 4 }}
  {{- end }}
```

unguiculus (changes requested):
Thanks.

charts/keycloak/templates/service-http.yaml
s-vkropotko: I use here `nindent` instead of `indent` so new line will be added with this function. If you want to follow suggested formatting I will change it to `indent`

charts/keycloak/Chart.yaml
s-vkropotko: updated

charts/keycloak/templates/service-http.yaml
s-vkropotko: updated

s-vkropotko:
Hi @unguiculus !
I've updated commit according to requested changes plz check it again

unguiculus (approved):

PR #357, opened 2020-12-14
[keycloak] fix: networking.k8s.io/v1/ingress check

jildertmiedema:
Checking only `.Capabilities.APIVersions.Has "networking.k8s.io/v1"` was not sufficiënt. Improved by checking `.Capabilities.APIVersions.Has "networking.k8s.io/v1/ingress"`.

The `backend` parts is also changed in v1, this is fixed now.

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 9.8.0
+version: 9.8.1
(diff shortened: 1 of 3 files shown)


unguiculus (approved):
Thanks

mteodori:
actually for me it works only with the former check, not the new one, I am using Docker Desktop mac which has v1.19.3, had to revert to 9.8.0

PR #359, opened 2020-12-15
[keycloak] Avoid issues with older K8s versions

### LINKED ISSUE #358: keycloak: OpenShift 3 compatibility broken
In #250, a workaround concerning `enableServiceLinks` was removed. `enableServiceLinks was only implemented in K8s 1.13 and thus is not compatible with OpenShift 3 (which is EOL sometimes 2021 but still widely used).

Should I create a PR fixing the issue?

mhutter:
enableServiceLinks was only introduced with K8s v1.13, which means the chart is incompatible with distributions that are still based on older versions (like OpenShift 3).

This commit changes the template to only render `enableServiceLinks` if it is enabled, allowing it the operator to disable it and restore pre-1.13 compatibility.

Fixes #358

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 9.8.1
+version: 9.8.2
 appVersion: 11.0.2
 description: Open Source Identity and Access Management For Modern Applications and Services
 keywords:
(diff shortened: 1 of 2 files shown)


srueg (approved):

unguiculus:
We've had a similar PR before (#320). I'm closing this for the same reason.

mhutter:
I'm very disappointed that features that in no way worsen the situation for current releases are rejected "just because".

Unfortunately in business environments commercially supported distributions are much more prevalent, and unfortunately they lag behind the vanilla K8s releases (sometimes _significantly_) while still being officially supported.

PR #360, opened 2020-12-20
[keycloak] fix networking.k8s.io/v1/Ingress API support

mteodori:
The fix introduced in #357 uses a lowercase check for the Ingress API `networking.k8s.io/v1/ingress` rather than `networking.k8s.io/v1/Ingress` which fails with a cluster version v1.19.3 like the one in the latest Docker Desktop.

This PR fixes that.

### DIFF
--- a/charts/keycloak/templates/ingress.yaml
+++ b/charts/keycloak/templates/ingress.yaml
@@ -35,7 +35,7 @@ spec:
         paths:
           {{- range .paths }}
           - path: {{ . }}
-            {{- if $.Capabilities.APIVersions.Has "networking.k8s.io/v1/ingress" }}
+            {{- if $.Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
             pathType: Prefix
             backend:
               service:
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 9.8.3
+version: 9.8.4
 appVersion: 11.0.2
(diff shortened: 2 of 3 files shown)


charts/keycloak/templates/ingress.yaml
```diff
@@ -35,13 +35,13 @@ spec:
...
           - path: {{ . }}
-            {{- if $.Capabilities.APIVersions.Has "networking.k8s.io/v1/ingress" }}
+            {{- if $.Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
             pathType: Prefix
             backend:
               service:
                 name: {{ include "keycloak.fullname" $ }}-http
                 port:
-                  name: {{ $ingress.servicePort }}
+                  {{ ternary "name" "number" (kindIs "string" $ingress.servicePort) }}: {{ $ingress.servicePort }}
```
unguiculus, line 44: I don't see why we want to change this. It's supposed to be a string. The named port should be referenced.

unguiculus (changes requested):

unguiculus (changes requested):
You will have to fix DCO.

mteodori:
> You will have to fix DCO.

done, couldn't find a way to do it from the GitHub UI though, also the conflict resolution...

unguiculus (approved):

mteodori:
thanks @unguiculus !

juliohm1978:
Has this been tested? It's not working for me. The chart never renders `apiVersion: networking.k8s.io/v1` because the Capability list does not contain `networking.k8s.io/v1/Ingress`. The list actually has `networking.k8s.io/v1`.

PR #362, opened 2021-01-08
Add dots after extraInitContainers section

hodjaev:
I added dots to help others get the example section right. When I used the keycloak's values file I didn't notice extraVolumeMounts and extraVolumes sections at the bottom of the file.


unguiculus:
Thanks for the PR, but I don't see how this is helpful. I prefer to not merge this.

PR #367, opened 2021-01-22
[mailhog] Add imagePullSecrets to chart

glothriel:
Due to this whole Docker quota drama we are mirroring all images to private registry, so we need this setting configurable.

Signed-off-by: Konstanty Karagiorgis <k.karagiorgis@piwik.pro>

### DIFF
--- a/charts/mailhog/templates/deployment.yaml
+++ b/charts/mailhog/templates/deployment.yaml
@@ -30,6 +30,10 @@ spec:
       securityContext:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       containers:
         - name: {{ template "mailhog.name" . }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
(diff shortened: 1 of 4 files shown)


glothriel:
Sorry @unguiculus for pinging, but It blocks the progress of what I'm currently doing - could You review this PR?

charts/mailhog/templates/deployment.yaml
```diff
@@ -30,6 +30,10 @@ spec:
       securityContext:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.image.pullSecrets }}
```
unguiculus, line 33: Don't group it under `image`.
```suggestion
      {{- with .Values.imagePullSecrets }}
```

unguiculus (changes requested):

charts/mailhog/templates/deployment.yaml
glothriel: Done. At this point, after configuring this setting in ~20 different helm charts I'm confident to say that I've seen almost every possible approach to setting this value :D

unguiculus: Yes, agreed. But this way is consistent with how it is in the manifest.

unguiculus (approved):

exaucae:
Bump: can't  we just apply this to the keycloak theme?

PR #369, opened 2021-01-28
[keycloak] Fix installation in numeric namespaces (#368)

### LINKED ISSUE #368: [keycloak] Chart can not be installed in numeric namespace
Keycloak chart installation with rbac fails if target namespace consists only of digits due to missing quoting in rbac definition.

hnicke:
Resolves #368 

Signed-off-by: Heiko Nickerl <dev@heiko-nickerl.com>

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 9.8.1
+version: 9.8.2
 appVersion: 11.0.2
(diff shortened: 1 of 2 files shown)


unguiculus (approved):
Why would one ever have namespaces that only contain digits?

hnicke:
At Flip GmbH, we automatically spin up a dev environment for every merge request. We use the pull request ID as namespace for the each dev environment. If I knew this would cause quite some pain along the road, we would have chosen another naming.

PR #370, opened 2021-01-28
[keycloak] Allow NodePort with LoadBalancer

marcin-je:
Similarly to Ingress-Nginx
(https://github.com/kubernetes/ingress-nginx/blob/master/charts/ingress-nginx/templates/controller-service.yaml#L40-L48)
allow to set NodePort when Service of type LoadBalancer is used.
The use case for this is to allow using specific node port when Network
Load Balancer is used in AWS setup and a specific port needs to be used
for a case where security groups need to be managed outside of
Kubernets.

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 9.8.2
+version: 9.9.0
 appVersion: 11.0.2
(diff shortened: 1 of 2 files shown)


unguiculus (dismissed):

unguiculus:
Please fix the conflict.

marcin-je:
@unguiculus all done

unguiculus (approved):

PR #373, opened 2021-02-01
[keycloak] Allow overriding app version label

zmingxie:
At the moment, the `app.kubernetes.io/version` label is always set to `.Chart.AppVersion`.

However, the keycloak app version string could also be overridden via `.Values.image.tag`. When this happens, the `app.kubernetes.io/version` will show an incorrect version string. This patch will update __helpers.tpl_ and allow it to also accept `.Values.image.tag` as the app version string if it is defined.

### DIFF
--- a/charts/keycloak/templates/_helpers.tpl
+++ b/charts/keycloak/templates/_helpers.tpl
@@ -37,9 +37,7 @@ Common labels
 {{- define "keycloak.labels" -}}
 helm.sh/chart: {{ include "keycloak.chart" . }}
 {{ include "keycloak.selectorLabels" . }}
-{{- if .Chart.AppVersion }}
-app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
-{{- end }}
+app.kubernetes.io/version: {{ .Values.image.tag | default .Chart.AppVersion | quote }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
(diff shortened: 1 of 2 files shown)


charts/keycloak/templates/_helpers.tpl
```diff
@@ -37,7 +37,9 @@ Common labels
 {{- define "keycloak.labels" -}}
 helm.sh/chart: {{ include "keycloak.chart" . }}
 {{ include "keycloak.selectorLabels" . }}
-{{- if .Chart.AppVersion }}
+{{- if .Values.image.tag }}
+app.kubernetes.io/version: {{ .Values.image.tag | quote }}
+{{- else }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
```
unguiculus, line 44: You could simplify that and follow the same pattern as for the Docker image configuration.
```suggestion
app.kubernetes.io/version: {{ .Values.image.tag | default .Chart.AppVersion | quote }}
{{- end }}
```

unguiculus (changes requested):
Makes sense. Thanks. I have on suggestion.

zmingxie:
@unguiculus Thanks for the suggestion! I've updated my patch with that.

unguiculus (approved):

PR #376, opened 2021-02-02
[keycloak] RBAC resource should include labels

zmingxie:
Even though we may not need to specify custom labels for RBAC resources, the `Role` and `RoleBinding` resource should still contain `keycloak.labels` like other resources created by this Helm chart


zmingxie:
@unguiculus Any comments on this small change?

unguiculus (approved):
Thanks @zmingxie. That was definitely an oversight. They shouldn't have missed from the start. 🎉

PR #387, opened 2021-02-15
fix extraEnv, extraEnvFrom json mappings and render method for it in …

v-chernov:
current values.json mapping of extraEnv and extraEnvFrom is String and it cannot be correctly mapped with Helm v3.5.1


v-chernov:
will reopen because of errors in linter

PR #391, opened 2021-02-15
[keycloak] fix extraEnv and extraEnvFrom mapping, values and deps

v-chernov:
- change extraEnv and extraEnvFrom to object in values.schema.json
- add install required postgres chart for PR linter

Signed-off-by: Vasily Chernov <vas.chernov@gmail.com>


unguiculus:
I'm sorry, this will not be changed. `tpl` is used on purpose. It is documented that way.

Issue #324, opened 2020-10-15
[keycloak] How to specify secrets's type: the type of secrets used in `imagePullSecrets` should be `kubernetes.io/dockercfg`


WaylandWong:
Hi, is there any way to specify the type of secret configured in the values file?
We can specify the `imagePullSecrets` for the Pod, but the secret should be pre-created. I want to configure this secret in the values file in 'secrets' section. Bug the secret type always be 'Opaque' while `imagePullSecrets` need a secret which type is 'kubernetes.io/dockercfg'.

I changed `templates/secrets` to like below and it works well. If this can help, I can create a pr for this issue.
```yaml
{{- range $nameSuffix, $values := .Values.secrets -}}
apiVersion: v1
kind: Secret
metadata:
  name: {{ include "keycloak.fullname" $ }}-{{ $nameSuffix }}
  {{- with $values.annotations }}
  annotations:
  {{- range $key, $value := . }}
  {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
  {{- end }}
  {{- end }}
  labels:
  {{- include "keycloak.labels" $ | nindent 4 }}
  {{- range $key, $value := $values.labels }}
  {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
  {{- end }}
{{- if $values.type }}
type: {{ $values.type }}
{{- else }}
type: Opaque
{{- end }}
{{- with $values.data }}
data:
  {{- toYaml . | nindent 2 }}
{{- end }}
{{- with $values.stringData }}
stringData:
  {{- range $key, $value := . }}
  {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 2 }}
  {{- end }}
{{- end }}
---
{{- end -}}

```

unguiculus:
Thanks for reporting this. Please create a PR as suggested.

Issue #40, opened 2019-05-24
Keycloak 5.0.0 Chart: specify  keycloak.persistence.deployPostgres: false and postgres is still deployed


enenum:
Was using keycloak 4.0 chart with with the above flawlessly and installing postgres separately.
With the new 5.0 chart, I end up with two deployments of postgres.

Helm & Tiller 2.11.0
Kubernetes 1.13 cluster installed with Kubespray 2.8

unguiculus:
There has been no change regarding the dependency. Please share your configuration so I can help.

unguiculus:
Closing due to inactivity.

rblaine95:
Hi there, I'm having the same issue over here with chart version `5.1.5`

```sh
$ helm version
Client: &version.Version{SemVer:"v2.12.1", GitCommit:"02a47c7249b1fc6d8fd3b94e6b4babf9d818144e", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.12.1", GitCommit:"02a47c7249b1fc6d8fd3b94e6b4babf9d818144e", GitTreeState:"clean"}

$ oc version
oc v3.11.0+0cbc58b
kubernetes v1.11.0+d4cacc0
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://localhost:8443
openshift v3.11.0+3b2d3b6-227
kubernetes v1.11.0+d4cacc0
```
```yaml
keycloak:
  persistence:
    deployPostgres: false
    dbVendor: postgres
    dbName: keycloak
    dbHost: kubedb-postgres
    dbPort: 5432
    dbUser: keycloak
    dbPassword: keycloak
```

Maybe I'm just missing something?

PR #398, opened 2021-02-19
[keycloak] Apply secret type configured in values file

WaylandWong:
specify secrets's type.
https://github.com/codecentric/helm-charts/issues/324

### DIFF
--- a/charts/keycloak/templates/secrets.yaml
+++ b/charts/keycloak/templates/secrets.yaml
@@ -14,7 +14,7 @@ metadata:
   {{- range $key, $value := $values.labels }}
   {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
   {{- end }}
-type: Opaque
+type: {{ default "Opaque" $values.type}}
 {{- with $values.data }}
 data:
   {{- toYaml . | nindent 2 }}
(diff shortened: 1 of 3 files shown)


charts/keycloak/templates/secrets.yaml
```diff
@@ -14,7 +14,11 @@ metadata:
   {{- range $key, $value := $values.labels }}
   {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
   {{- end }}
+{{- if $values.type }}
+type: {{ $values.type }}
+{{- else }}
 type: Opaque
+{{- end }}
```
unguiculus, line 21: I'd suggest you simply this:
```suggestion
type: {{ default $values.type "Opaque" }}
```

unguiculus (changes requested):
Thanks. Please fix DCO and bump the chart version.

Also, please update the comment in the values file:
https://github.com/codecentric/helm-charts/blob/master/charts/keycloak/values.yaml#L209-L214

charts/keycloak/templates/secrets.yaml
WaylandWong: Good idea. And I think it would be 
```
type: {{ default "Opaque" $values.type}}
```

WaylandWong:
I change the version to `9.9.5`, so please merge #40 first.

charts/keycloak/templates/secrets.yaml
unguiculus: You're right, of course 😉

unguiculus (approved):

PR #372, opened 2021-01-28
[keycloak] Install RBAC in namespace, not cluster-wide

hnicke:
Resolves #371 

Signed-off-by: Heiko Nickerl <dev@heiko-nickerl.com>

PR #401, opened 2021-02-23
[keycloak] Add env var in keycloak chart

cuttingedge1109:
jgroups-kubernetes performs KUBE_PING in KUBERNETES_NAMESPACE namespace.
If not specified, then it performs against default namespace.
We can set `KUBERNETES_NAMESPACE` env var in `extraEnv` but I think auto-set is more reasonable.

Before #372, clusterRole was provisioned so no need to set the `KUBERNETES_NAMESPACE` explicitly.
But now, need to specify it because role and rolebinding is provisioned.

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 9.9.5
+version: 9.9.6
 appVersion: 11.0.3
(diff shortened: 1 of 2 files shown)


unguiculus:
Thanks for the PR. This is already possible without adding an explicit environment variable to the chart. Please have a look at the readme. Your use case is documented at https://github.com/codecentric/helm-charts/tree/master/charts/keycloak#kube_ping-service-discovery

PR #402, opened 2021-02-23
[keycloak] Adding externalTrafficPolicy to service definitions

robcoward:
This PR allows the deployment to explicitly set the `externalTrafficPolicy` when deploying a service type 'LoadBalancer', allowing the default value of `Cluster` to be changed to `Local` as part of the chart provisioning of the load balanced service.


charts/keycloak/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 9.9.5
+version: 9.9.6
```
unguiculus, line 3: Please bump the minor version for a new feature to follow semver.
```suggestion
version: 9.10.0
```

charts/keycloak/values.schema.json
```diff
@@ -431,4 +431,4 @@
       }
     }
   }
-}
+}
```
unguiculus, line 434: Keep the line break
```suggestion
}

```

charts/keycloak/templates/service-http.yaml
```diff
@@ -23,6 +23,9 @@ spec:
   loadBalancerSourceRanges:
     {{- toYaml .Values.service.loadBalancerSourceRanges | nindent 4 }}
   {{- end }}
+  {{- if and (eq "LoadBalancer" .Values.service.type) .Values.service.externalTrafficPolicy }}
+  externalTrafficPolicy: {{ .Values.service.loadBalancerSourceRanges }}
+  {{- end }}
```
unguiculus, line 28: Move default value to values.yaml
```suggestion
  {{- if eq "LoadBalancer" .Values.service.type }}
  externalTrafficPolicy: {{ .Values.service.loadBalancerSourceRanges }}
  {{- end }}
```

charts/keycloak/values.yaml
```diff
@@ -242,6 +242,10 @@ service:
   # to connect to the LoadBalancer, e. g. will result in Security Groups
   # (or equivalent) with inbound source ranges allowed to connect
   loadBalancerSourceRanges: []
+  # When using Service type LoadBalancer, you can preserve the source IP seen in the container
+  # by changing the default (Cluster) to be Local.
+  # See https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip
+  externalTrafficPolicy: ""
```
unguiculus, line 248: Add default value here
```suggestion
  externalTrafficPolicy: Cluster
```

unguiculus (changes requested):
Please update the readme as well.

charts/keycloak/Chart.yaml
robcoward: minor version bumped as requested

charts/keycloak/templates/service-http.yaml
robcoward: Updated as requested

charts/keycloak/values.yaml
robcoward: Updated as requested

PR #408, opened 2021-03-12
[keycloak] Add startupProbe

as3445:
Since Keycloak takes a significant amount of time to start up, it is helpful to have startupProbe to find out when it's ready to receive connections.

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 10.1.0
+version: 10.1.1
 appVersion: 12.0.4
 description: Open Source Identity and Access Management For Modern Applications and Services
 keywords:
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 10.1.1
+version: 10.2.0
 appVersion: 12.0.4
 description: Open Source Identity and Access Management For Modern Applications and Services
 keywords:
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 10.2.0
+version: 10.3.0
 appVersion: 12.0.4
 description: Open Source Identity and Access Management For Modern Applications and Services
 keywords:
--- a/charts/keycloak/templates/statefulset.yaml
+++ b/charts/keycloak/templates/statefulset.yaml
@@ -127,6 +127,10 @@ spec:
           readinessProbe:
             {{- tpl . $ | nindent 12 }}
           {{- end }}
+          {{- with .Values.startupProbe }}
+          startupProbe:
+            {{- tpl . $ | nindent 12 }}
+          {{- end }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
           volumeMounts:
--- a/charts/keycloak/values.yaml
+++ b/charts/keycloak/values.yaml
@@ -166,6 +166,14 @@ readinessProbe: |
   initialDelaySeconds: 30
   timeoutSeconds: 1
 
+# Startup probe configuration
+startupProbe: |
+  httpGet:
+    path: /auth/realms/master
+    port: http
+  initialDelaySeconds: 300
+  timeoutSeconds: 1
+
 # Pod resource requests and limits
 resources: {}
   # requests:
--- a/charts/keycloak/values.yaml
+++ b/charts/keycloak/values.yaml
@@ -173,7 +173,7 @@ startupProbe: |
     port: http
   initialDelaySeconds: 30
   timeoutSeconds: 1
-  failureThreshold: 30
+  failureThreshold: 60
   periodSeconds: 5
 
 # Pod resource requests and limits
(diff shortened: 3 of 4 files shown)


charts/keycloak/templates/statefulset.yaml
```diff
@@ -127,6 +127,10 @@ spec:
           readinessProbe:
             {{- tpl . $ | nindent 12 }}
           {{- end }}
+          {{- with .Values.readinessProbe }}
```
as3445, line 130:
```suggestion
          {{- with .Values.startupProbe }}
```

charts/keycloak/templates/statefulset.yaml
```diff
@@ -127,6 +127,10 @@ spec:
...
+          readinessProbe:
```
as3445, line 131:
```suggestion
          startupProbe:
```

charts/keycloak/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 10.1.0
+version: 10.1.1
```
thomasdarimont, line 3: Please bump the minor version for a new feature

```suggestion
version: 10.2.0
```

thomasdarimont (changes requested):
Thanks for the contribution! Please see the comments inline.

charts/keycloak/values.yaml
```diff
@@ -166,6 +166,14 @@ readinessProbe: |
   initialDelaySeconds: 30
   timeoutSeconds: 1
 
+# Startup probe configuration
+startupProbe: |
+  httpGet:
+    path: /auth/realms/master
```
unguiculus, line 172: A startup probe complements the liveness probe. The liveness probe only kicks in once the startup probe has succeeded. It should, thus, use the same path as the liveness probe.

```suggestion
    path: /auth/
```

Along with that, the `initialDelaySeconds` of the liveness probe should be reduced. There's no point in having a huge delay anymore. I guess we can set it to 0.

unguiculus (changes requested):

as3445:
Not sure if the chart tests are flaky or if there's an issue with my PR, help would be appreciated!

unguiculus:
Again please adjust the `initialDelaySeconds` of the liveness probe as suggested.

as3445:
> Again, please adjust the `initialDelaySeconds` of the liveness probe as suggested. CI failed due to a timeout. Might be related to the long delay in the liveness probe.

Lowered that as well as the value for the readiness probe. Does not seem to help the issue. 🤔

as3445:
Halving the initialDelaySeconds of the startupProbe seems to have done the trick, let me know if the numbers all look good to everyone. Happy to adjust it is necessary.

charts/keycloak/values.yaml
```diff
@@ -155,17 +155,25 @@ livenessProbe: |
   httpGet:
     path: /auth/
     port: http
-  initialDelaySeconds: 300
+  initialDelaySeconds: 5
```
unguiculus, line 158:
```suggestion
  initialDelaySeconds: 0
```

charts/keycloak/values.yaml
```diff
@@ -155,17 +155,25 @@ livenessProbe: |
...
   timeoutSeconds: 5
 
 # Readiness probe configuration
 readinessProbe: |
   httpGet:
     path: /auth/realms/master
     port: http
-  initialDelaySeconds: 30
+  initialDelaySeconds: 5
```
unguiculus, line 166: Reducing this doesn't make much sense. Keep it as it was.
```suggestion
  initialDelaySeconds: 30
```

charts/keycloak/values.yaml
```diff
@@ -155,17 +155,25 @@ livenessProbe: |
...
   timeoutSeconds: 1
 
+# Startup probe configuration
+startupProbe: |
+  httpGet:
+    path: /auth/
+    port: http
+  initialDelaySeconds: 150
+  timeoutSeconds: 5
```
unguiculus, line 175: This will slow down startup a lot because it waits 150 s until the probe even kicks in, no matter how fast startup really is. Instead, we should increase the failure threshold.

```suggestion
  initialDelaySeconds: 30
  timeoutSeconds: 1
  failureThreshold: 30
  periodSeconds: 5
```

unguiculus (changes requested):
I have a few more remarks. Please make sure you update the readme as well.

as3445:
I think I've addressed all the comments! 🙏

charts/keycloak/values.yaml
```diff
@@ -166,6 +166,16 @@ readinessProbe: |
...
   timeoutSeconds: 1
 
+# Startup probe configuration
+startupProbe: |
+  httpGet:
+    path: /auth/
+    port: http
+  initialDelaySeconds: 30
+  timeoutSeconds: 1
+  failureThreshold: 30
```
mircohaug, line 176:
```suggestion
  failureThreshold: 60
```
This should be 60 to keep the 300 second timeout from before.

charts/keycloak/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 10.1.0
+version: 10.2.0
```
mircohaug, line 3: Another PR was merged in between and you have to increase the version number once again... Sorry for that

mircohaug (changes requested):
Sorry for the long wait. I have some minor comments but afterwards it looks good to me. @thomasdarimont is there still anything you want changed?

charts/keycloak/Chart.yaml
as3445: No problem, I've incorporated your feedback!

mircohaug (approved):
LGTM!

PR #409, opened 2021-03-17
[mailhog] Add Outgoing SMTP config

reixd:
Add config to add permanent outgoing-smtp config using a secret analogue to the auth secret.


mircohaug (approved):
Thanks for the PR. Looks good to me

PR #410, opened 2021-03-18
[keycloak] update postgres subchart to 10.3.13

### LINKED ISSUE #336: [Keycloak] DB initialization & migration
Currently the database migration of Keycloak happens while the pod starts. However if an healthcheck is added with an initialdelay of let's say 10 seconds the deployment of keycloak will fail. This is due to not being live at that moment of time. Is there any chance initialization and migration will be added as separate hooks which can be ran during pre-upgrade phase? This way the hook will fire first (without any healthchecks going on) so the application can correctly migrate the db. I'd propose 2 hooks, one for initializing and the other for seeding the database.

jonkerj:
Fixes #336

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 10.3.0
+version: 11.0.0
 appVersion: 12.0.4
 description: Open Source Identity and Access Management For Modern Applications and Services
 keywords:
--- a/charts/keycloak/templates/_helpers.tpl
+++ b/charts/keycloak/templates/_helpers.tpl
@@ -65,7 +65,7 @@ Create a default fully qualified app name for the postgres requirement.
 */}}
 {{- define "keycloak.postgresql.fullname" -}}
 {{- $postgresContext := dict "Values" .Values.postgresql "Release" .Release "Chart" (dict "Name" "postgresql") -}}
-{{ include "postgresql.fullname" $postgresContext }}
+{{ include "postgresql.primary.fullname" $postgresContext }}
 {{- end }}
 
 {{/*
--- a/charts/keycloak/README.md
+++ b/charts/keycloak/README.md
@@ -646,6 +646,16 @@ The headless service that governs the StatefulSet is used for DNS discovery via 
 
 ## Upgrading
 
+### From chart < 11.0.0
+
+When you are using the postgresql subchart (which is not recommended for production situations), you will need to migrate the statefulset governing the postgresql pod.
+The upgrade will cause some immutable fields of the statefulset to be modified, which cannot be done in place.
+You will have to manually remove the statefulset before doing the Helm upgrade.
+The following procedure takes care of this:
+
+1. Remove the old statefulset (associated PVC will remain in place): `kubectl delete statefulset -n <your ns> <name of your psql statefulset>`
+1. Helm Upgrade: `helm upgrade -n <your ns> ... <your release>`
+
 ### From chart < 10.0.0
 
 * Keycloak is updated to 12.0.4
--- a/charts/keycloak/requirements.lock
+++ b/charts/keycloak/requirements.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: postgresql
   repository: https://charts.bitnami.com/bitnami
-  version: 9.1.1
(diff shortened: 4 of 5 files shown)


jonkerj:
I somehow doubt the timeout from the lint CI job is caused by my code, will it retry automatically?

jonkerj:
Dear maintainers, I'd like some feedback on this PR. The old situation carries a (very) old Postgres, which lacks functionality (startupProbe) and is a possible security risk. It has to be updated at some point.

charts/keycloak/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 10.1.0
+version: 10.2.0
```
mircohaug, line 3: Please update the chart version

charts/keycloak/templates/_helpers.tpl
```diff
@@ -65,7 +65,7 @@ Create a default fully qualified app name for the postgres requirement.
 */}}
 {{- define "keycloak.postgresql.fullname" -}}
 {{- $postgresContext := dict "Values" .Values.postgresql "Release" .Release "Chart" (dict "Name" "postgresql") -}}
-{{ include "postgresql.fullname" $postgresContext }}
+{{ include "postgresql.primary.fullname" $postgresContext }}
```
mircohaug, line 68: on local install I get the error message. Please fix this or show me what i did wrong.
```
❯ helm install my-test4 .
Error: template: keycloak/templates/statefulset.yaml:58:48: executing "keycloak/templates/statefulset.yaml" at <include "keycloak.postgresql.fullname" .>: error calling include: template: keycloak/templates/_helpers.tpl:68:3: executing "keycloak.postgresql.fullname" at <include "postgresql.primary.fullname" $postgresContext>: error calling include: template: no template "postgresql.primary.fullname" associated with template "gotpl"
```

mircohaug (changes requested):
Thanks for the PR. 

Just for your Information: In productive setups we do not recommend to use the build-in (subchart) postgres but instead maintain a dedicated database separately.

charts/keycloak/Chart.yaml
jonkerj: Bumped version and force-pushed.

charts/keycloak/templates/_helpers.tpl
jonkerj: That's a hard one, since I cannot reproduce your error:
```
$ git status
On branch bump-postgres-in-keycloak
$ helm install my-test42 .
NAME: my-test42
LAST DEPLOYED: Thu Apr 15 20:19:59 2021
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
***********************************************************************
*                                                                     *
*                Keycloak Helm Chart by codecentric AG                *
*                                                                     *
***********************************************************************

Keycloak was installed with a Service of type ClusterIP

Create a port-forwarding with the following commands:

export POD_NAME=$(kubectl get pods --namespace default -l "app.kubernetes.io/name=keycloak,app.kubernetes.io/instance=my-test42" -o name)
echo "Visit http://127.0.0.1:8080 to use your application"
kubectl --namespace default port-forward "$POD_NAME" 8080
```
```

mircohaug: My bad, i forgot to run `helm dep build`. 

But the old and the new version are not compatible out of the box. When upgrading an old release I get the error `Error: UPGRADE FAILED: cannot patch "test-postgresql" with kind StatefulSet: StatefulSet.apps "test-postgresql" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden`

This means this change requires a major version upgrade of the chart as well as upgrade instructions. In this case typically a `kubectl delete --cascade=false sts test-postgresql`, `helm upgrade`, ` kubectlk delete olad database pods`

jonkerj: Thanks for your feedback and testing.

This is resolved per my latest force push. Your procedure (delete without cascade) is (IMHO) more complex than necessary: just deleting the STS will keep the PVC in place. This is what most helm charts I use recommend as upgrade path in this situation, also.

mircohaug (approved):

mircohaug:
Thanks for your work and energy. Now it is done ;-)

PR #413, opened 2021-03-30
[keycloak] Adding Node Selector to Test Pod

behrad84:
Hello codecentric, 

I think it is good idea to modify Test Pod definition to include NodeSelector.


behrad84:
closing this PR.

PR #416, opened 2021-04-05
[keycloak] Adding externalTrafficPolicy to service definitions

robcoward:
This PR allows the deployment to explicitly set the externalTrafficPolicy when deploying a service type 'LoadBalancer', allowing the default value of Cluster to be changed to Local as part of the chart provisioning of the load balanced service.


mircohaug (approved):
Thanks for the PR. LGTM

PR #417, opened 2021-04-16
[mailhog] Fix to enable outgoingSMTP and auth

quorak:
Currently outgoingSMTP and auth cannot be enabled at the same time, as the keys are overwriting themselves. THis MR should fix it.


mircohaug:
Have a look at the DCO bot. Your commit have to be signed off in order to be allowed. Disclaimer: No crypto involved. This simple adds a line to your commit messages.

PR #432, opened 2021-06-07
[keycloak] Fix container ports in network policy

ccremer:
Network policies operate on the pod level, not the service. 
Allowing (previous default) port 80 from service doesn't work, since container listens on 8080.

(Since the container port isn't configurable in the values.yaml, I also hardcoded those in the netpol template.)


mircohaug (approved):
Good catch. LGTM

PR #442, opened 2021-07-06
[keycloak] Allow templating in prometheus rules

jBouyoud:
Most of monitoring rules are namespace dependent.
When we have two instances of keycloak chart on one cluster and we want to share our monitoring rules across these chart instances, we need to have templating capabilities.

Signed-off-by: Julien Bouyoud <jBouyoud@users.noreply.github.com>

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 14.0.1
+version: 14.1.0
 appVersion: 14.0.0
 description: Open Source Identity and Access Management For Modern Applications and Services
 keywords:
--- a/charts/keycloak/templates/prometheusrule.yaml
+++ b/charts/keycloak/templates/prometheusrule.yaml
@@ -19,6 +19,14 @@ spec:
   groups:
     - name: {{ include "keycloak.fullname" $ }}
       rules:
+        {{- with .rules }}
         {{- toYaml .rules | nindent 8 }}
+        {{- end }}
+        {{- with .rulesTemplate }}
+        {{- range . }}
+        -
+          {{- (tpl (. | toYaml) $) | nindent 10 }}
+        {{- end }}
+        {{- end }}
 {{- end }}
 {{- end -}}


jBouyoud:
Kind cluster setup seems flaky : ` • WARNING: Timed out waiting for Ready ⚠️`

mircohaug:
This change would potentially break all existing deployments of the chart because the tpl function can only be used on strings and currently the prometheus rules are objects. 

Can you please describe your usecase which requires this change in higher detail? As I understood your use case this could be achieved by defining no rule in the values of instance 1 and add the rule to target both to the values of instance 2. Please correct me if I'm wrong.

jBouyoud:
Hi, 

You right, I didn't see that. To avoid the breaking change we may use another configuration key like `rulesTemplate`. 

_My use case:_
I've a SSO chart based on Keycloak and others dependencies, when I install my chart I want to provide Keycloak fully configured for our environment including monitoring rules (and many others stuff like grafana dashboard,....). 
My SSO chart is installed many time on our K8s cluster, namespace isolated (looks pretty standard). 
And I don't want to copy paste, all monitoring rules configuration each times I need a new stack (which could occurs many time a day). 
The intent beside this is to provide at our organization level a well configured chart that can't be provided by a public chart (due to our policies, conventions, ...).

Exemple : 
In my chart `values.yaml` I'd like to write the following alerting rule : 
```
keycloak:
  prometheusRule:
    enabled: true
    labels: 
       <<: *our-organisation-conventionnal-labels-ref
    metricsInstanceSelector: |-
      {namespace="{{ .Release.Namespace }}", app_kubernetes_io_name="{{ include "keycloak.fullname" . }}"}
    rules:
       -  alert: TooHighRegistrationFailure
          expr: |
            irate(keycloak_registrations_errors{{ tpl .Values.metrics.prometheusRule.metricsInstanceSelector . }}) > 0
          for: 2m
          labels:
            severity: warn
            <<: *our-prometheus-conventionnal-labels-ref
         annotations:
           description: "An important......"
           summary: "My summary.... "
           runbook_url: 'https://github.com/company/sso/OPERATION.md#TooHighRegistrationFailure'
```

Pleas let me known if you have other questions.

jBouyoud:
Hi @mircohaug,

I've updated this PR to remove the breaking changes. You can also find my use case explain's in the previous comment.
I hope this helps. I remain available to handle any changes ;-P

PR #447, opened 2021-07-19
[keycloak] Fix #446 by making pathType for ingress rules configurable

### LINKED ISSUE #446: [Keycloak] Make pathType configurable for ingress
In Google Kubernetes Engine (GKE) the only supported `pathType` for ingresses is `ImplementationSpecific`. Currently the keycloak helm chart only supports the value `Prefix`, which does not work on GKE.

So it would be nice, if we could configure the `pathType` to our own needs.

cedosw:
Signed-off-by: Cedric Oswald <c.oswald@hmstr.de>

### DIFF
--- a/charts/keycloak/templates/ingress.yaml
+++ b/charts/keycloak/templates/ingress.yaml
@@ -34,9 +34,9 @@ spec:
       http:
         paths:
           {{- range .paths }}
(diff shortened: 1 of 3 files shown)


thomasdarimont:
Superseded by #450

PR #453, opened 2021-07-20
[keycloak] separated console tls hosts and secrets

alirezaDavid:
Hey
It might be a good idea if we can separate TLS hosts and secrets for the console ingress.

Signed-off-by: Alireza Davoodi <davoodi@fd.nl>


Bouwdie:
This is needed when you want to run the console under a different host. @mircohaug, @thomasdarimont it looks like the separate console ingress feature is not designed for this, correct? What would be a good way to support this you think? :1st_place_medal:

PR #457, opened 2021-07-26
[keycloak] fix http port in case of reencrypt of passthrough

cegiraud:
The port stays to "http" in case of reencrypt of passthrough, fixed that


cegiraud:
In case of route reencryp or passthrough, the targetPort stays "http" and not "https".

thomasdarimont:
@cegiraud Thanks for your contribution! Would you mind rebasing this again?

Also please prefix your commit message with [keycloak]

mircohaug (approved):

PR #458, opened 2021-07-26
[keycloak] add a new scope extraPort in extraFrom for NetworkPolicies

cegiraud:
Signed-off-by: GIRAUD Charles-Eric <charles-eric.giraud@soprasteria.com>


cegiraud:
Hi,

In the extra from we don't want to expose all ports defined in extraPort (like JGroups for example).

PR #462, opened 2021-08-04
[keycloak] Support for SQLServer added, initContainers for DB and User creation added to.

mdepa:
Updated the original helm chart to provide explicit support to SQL Server.

Added capability to automated database and login creation.

### DIFF
--- a/charts/keycloak/values.yaml
+++ b/charts/keycloak/values.yaml
@@ -62,12 +62,12 @@ rbac:
 # SecurityContext for the entire Pod. Every container running in the Pod will inherit this SecurityContext. This might be relevant when other components of the environment inject additional containers into running Pods (service meshes are the most prominent example for this)
 podSecurityContext:
   fsGroup: 1000
-
+  # runAsGroup: 1000730000
 # SecurityContext for the Keycloak container
 securityContext:
   runAsUser: 1000
   runAsNonRoot: true
-
+  # runAsGroup: 1000730000
 # Additional init containers, e. g. for providing custom themes
 extraInitContainers: ""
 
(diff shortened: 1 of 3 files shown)


charts/keycloak/values.yaml
```diff
@@ -62,12 +62,12 @@ rbac:
 # SecurityContext for the entire Pod. Every container running in the Pod will inherit this SecurityContext. This might be relevant when other components of the environment inject additional containers into running Pods (service meshes are the most prominent example for this)
 podSecurityContext:
   fsGroup: 1000
-
+  # runAsGroup: 1000730000
```
thomasdarimont, line 65: Do we need this comment?

charts/keycloak/values.yaml
```diff
@@ -374,6 +381,26 @@ postgresql:
   networkPolicy:
     enabled: false
 
+mssql:
+  # secret containing the sa password
+  existingSaSecret: sqlserver-secret
+  existingSaSecretKey: sapassword
+  # If `true`, keycloak will use the SQL Server DB engine
+  enabled: true
```
thomasdarimont, line 389: I'd move the `enabled` flag up as the first attribute.

charts/keycloak/values.yaml
```diff
@@ -363,7 +370,7 @@ pgchecker:
 
 postgresql:
   # If `true`, the Postgresql dependency is enabled
-  enabled: true
+  enabled: false
```
thomasdarimont, line 373: I'd keep postgresql as the default and disable the sqlserver dependency by default.

thomasdarimont (changes requested):

Issue #459, opened 2021-07-27
extraEnv substitution (feature request)


hchan:
I want the ability to have the extraEnv variable merge between to values files.
i.e. if I have values.yaml and values-secrets.yaml, I want values-secrets.yaml to merge certain PASSWORDs to values.yaml

beastob:
May I confirm if the following scenario is what you expected.
Let's assume the default values.yaml contains:
```yaml
## values.yaml 
extraEnv:
  - name: KEYCLOAK_LOGLEVEL
    value: DEBUG
  - name: WILDFLY_LOGLEVEL
    value: DEBUG
  - name: CACHE_OWNERS_COUNT
    value: "2"
  - name: CACHE_OWNERS_AUTH_SESSIONS_COUNT
    value: "2"
```
And your own custom values to override
```yaml
## values-custom.yaml
extraEnv:
  - name: KEYCLOAK_LOGLEVEL
    value: INFO
  - name: WILDFLY_LOGLEVEL
    value: INFO
  - name: PASSOWRD
    value: 'changeme'
  - name: USERNAME
    value: admin
```
The result combined values for the Helm chart would be:
```yaml
## combined values
extraEnv:
  - name: KEYCLOAK_LOGLEVEL
    value: INFO
  - name: WILDFLY_LOGLEVEL
    value: INFO
  - name: CACHE_OWNERS_COUNT
    value: "2"
  - name: CACHE_OWNERS_AUTH_SESSIONS_COUNT
    value: "2"
  - name: PASSOWRD
    value: 'changeme'
  - name: USERNAME
    value:  #admin
```

hchan:
Yes sir, that's what I want.  Right now extraEnv is a String rather than a hash

ccremer:
Hi!

We have a very similar problem.

See an example of the problem here: https://github.com/projectsyn/component-keycloak/blob/783e77cd4d290a3ee7909cb5537b31dd11c41555/class/defaults.yml#L135

For context:
We are wrapping the Keycloak Helm chart in a Project Syn component. In short: Project Syn enables Puppet Hiera like configuration hierarchy through YAML. In this hierarchy, we can pull parameters from multiple sources (e.g. global configuration down to cluster-specific configuration) and merge them. Deep-Merging best works with hashes/dicts/objects, but not with strings or arrays.

PR #466, opened 2021-08-19
[keycloak] extraEnv substitution (feature request) #459

beastob:
## Convert previous parameters which take in multi-line `string` instead of `array` or `object`
Result example:
```yaml
extraEnv:
  - name: KEYCLOAK_LOGLEVEL
    value: DEBUG
  - name: WILDFLY_LOGLEVEL
    value: DEBUG
```
Such change will allow partial override of `extraEnv`, where currently `extraEnv: |` will have to override all of its child values.

## Changes apply to the following parameters:
- livenessProbe
- readinessProbe
- startupProbe
- extraInitContainers
- extraContainers
- lifecycleHooks
- extraEnv
- extraEnvFrom
- extraVolumes
- extraVolumeMounts

## **Known Issues (Two so far):**

1. Parameter `affinity` is left unchanged due the an error caused by
  ```yaml
   `{{- include "keycloak.selectorLabels" . | nindent 10 }}`
  ```
  > Error: cannot load values.yaml: error converting YAML to JSON: yaml: line 120: did not find expected node content

2. After the changes the values from template are missing the single-quote `' '`
  Input values:
  ```yaml
  - name: JGROUPS_DISCOVERY_PROPERTIES
    value: 'dns_query={{ include "keycloak.serviceDnsName" . }}'
  ```
   Outcome after update:
  ```yaml
  - name: JGROUPS_DISCOVERY_PROPERTIES
    value: dns_query=RELEASE-NAME-keycloa-headless.default.svc.cluster.local
  ```
  
   Original outcome:
  ```yaml
   - name: JGROUPS_DISCOVERY_PROPERTIES
     value: 'dns_query=RELEASE-NAME-keycloa-headless.default.svc.cluster.local'
  ```

**Please comment below if you have any concerns or suggestions about the changes.**

### DIFF
--- a/charts/keycloak/README.md
+++ b/charts/keycloak/README.md
@@ -68,8 +68,8 @@ The following table lists the configurable parameters of the Keycloak chart and 
 | `clusterDomain` | The internal Kubernetes cluster domain | `cluster.local` |
 | `command` | Overrides the default entrypoint of the Keycloak container | `[]` |
 | `args` | Overrides the default args for the Keycloak container | `[]` |
-| `extraEnv` | Additional environment variables for Keycloak | `""` |
-| `extraEnvFrom` | Additional environment variables for Keycloak mapped from a Secret or ConfigMap | `""` |
+| `extraEnv` | Additional environment variables for Keycloak | `[]` |
+| `extraEnvFrom` | Additional environment variables for Keycloak mapped from a Secret or ConfigMap | `[]` |
 | `priorityClassName` | Pod priority class name | `""` |
 | `affinity` | Pod affinity | Hard node and soft zone anti-affinity |
 | `nodeSelector` | Node labels for Pod assignment | `{}` |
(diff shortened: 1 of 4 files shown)


ccremer:
Hi!
Any chance this gets some attention for a review?

I do have the idea to make those parameters not an array, but rather objects. The reason for objects is basically that it makes those parameters deep-mergeable when using Helm-wrapping tools like [Helmfile](https://github.com/roboll/helmfile) or [Kapitan](https://kapitan.dev/)

Usage example:

```yaml
extraEnv:
  someIrrelevantKey:
    name: JGROUPS_DISCOVERY_PROPERTIES
    value: dns_query=RELEASE-NAME-keycloa-headless.default.svc.cluster.local
```
The `someIrrelevantKey` is only there for enabling deep-merging and is being consumed by a for-iterator. It could be named the same as `JGROUPS_DISCOVERY_PROPERTIES` in the example.
Implementation could look like
```
            {{- with .Values.extraEnv }}
            {{ range $key, $value := . }}
            - {{- tpl (toYaml $value) $ | nindent 12 }}            
            {{- end }}
            {{- end }}
```
(I didn't check if the code sample would work)

Your thoughts?

mircohaug:
Hi thanks for waiting this long ;-) a lot of these values are strings by design. The reason is that you can't do for loops in your approach or in more abstract terms: It is not possible to use control flow elements in your approach. 

But you are correct this effectively prevents deep merging of values especially in the presence of higher order tools. 

I would rather not merge this in the current state (which would effectively remove functionality) but I would be open to a PR which keeps the freedom in a set of overwrite fields and migrate the defaults like you proposed them

PR #472, opened 2021-08-31
[keycloak] fix uncut version label

eumel8:
We use sha256 image tags which are longer then allowed label length and caused a deployment error:
```
Invalid value: "52771198613a47a3c3a617937c137a61aa6591267ba6ee03f63ff04b61d5f3f8": must be no more than 63 characters
```
truncate the label after 63 characters.


mircohaug (dismissed):

mircohaug (approved):

thomasdarimont (approved):
LGTM

PR #481, opened 2021-09-15
[mailhog] support ingressClassName on ingress

hypnopotamus:
ingressClassName is the preferred way to associated Ingress resources with their controller, the annotation is deprecated (https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation)

the annotation should still be respected by ingress controllers (https://pkg.go.dev/k8s.io/api/networking/v1#IngressSpec: "IngressClassName is the name of the IngressClass cluster resource. The associated IngressClass defines which controller will implement the resource. This replaces the deprecated `kubernetes.io/ingress.class` annotation. For backwards compatibility, when that annotation is set, it must be given precedence over this field...")

### DIFF
--- a/charts/mailhog/Chart.yaml
+++ b/charts/mailhog/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 description: An e-mail testing tool for developers
 name: mailhog
 appVersion: v1.0.1
-version: 5.0.0
+version: 5.0.1
 type: application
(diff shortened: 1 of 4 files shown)


hypnopotamus:
same as #477 applied to mailhog's chart

mircohaug (approved):
LGTM Thanks for your contribution

PR #488, opened 2021-09-29
[keycloak] fix(_helpers.tpl): redefine `keycloak.postgresql.fullname` to use postgresql replication setup

vmaillot:
While testing the replication setup in postgresql with keycloak I found a bug in the chart that will occur in the `pgchecker` `initContainer` while waiting for the postgresql server to be up and running.

Indeed when using `postgresql.primary.fullname` we ended up having  `keycloak.postgresql.fullname` being set to `RELEASE-NAME-postgresql-primary` because of these [lines](https://github.com/bitnami/charts/blob/e34ee5c5aeafabe15cbf0d2b0fecb18483bc6415/bitnami/postgresql/templates/_helpers.tpl#L17-L18). Therefore the `pgchecker` container would never resolve `RELEASE-NAME-postgresql-primary` because no such service is deployed while using postgresql replication setup and just be stuck on `Init:0/1` because of hanging `nc -z -w 2 keycloak-postgresql-primary 5432` DB server check command.

I tested this fix with and without replication set, it works with both setup now.

Thanks for the concern of this fix.

Tell me if I am missing anything!

Signed-off-by: Valentin Maillot <valentin.maillot@adfinis.com>

### DIFF
--- a/charts/keycloak/templates/_helpers.tpl
+++ b/charts/keycloak/templates/_helpers.tpl
@@ -65,7 +65,7 @@ Create a default fully qualified app name for the postgres requirement.
 */}}
 {{- define "keycloak.postgresql.fullname" -}}
 {{- $postgresContext := dict "Values" .Values.postgresql "Release" .Release "Chart" (dict "Name" "postgresql") -}}
-{{ include "keycloak.name" .}}-{{ include "postgresql.name" $postgresContext }}
+{{ include "keycloak.fullname" .}}-{{ include "postgresql.name" $postgresContext }}
 {{- end }}
 
 {{/*
--- a/charts/keycloak/templates/_helpers.tpl
+++ b/charts/keycloak/templates/_helpers.tpl
@@ -65,7 +65,7 @@ Create a default fully qualified app name for the postgres requirement.
 */}}
 {{- define "keycloak.postgresql.fullname" -}}
 {{- $postgresContext := dict "Values" .Values.postgresql "Release" .Release "Chart" (dict "Name" "postgresql") -}}
-{{ include "postgresql.primary.fullname" $postgresContext }}
+{{ include "keycloak.name" .}}-{{ include "postgresql.name" $postgresContext }}
 {{- end }}
 
 {{/*


vmaillot:
The CI is failing due to the deployment with `postgresql-ha` instead of `postgresql` helm chart. Why are you doing this? Because I the keycloak chart itself I only see `postgresql` as dependency and not `postgresql-ha`. So that makes total sense that the CI is failing because in the end there is no service deployed named `keycloak-postgresql`.

Thanks for you reply!

vmaillot:
@thomasdarimont @mircohaug

vmaillot:
@thomasdarimont @mircohaug could you please re-evaluate this PR and fix? Please see issue [#620].

eyenx:
This is the problem we also face: 

```
/tmp/keycloak helm template keycloak . --set postgresql.replication.enabled=true --set postgresql.persistence.enabled=true | grep Service$ -A 3
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /home/eye/.kube/config
WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /home/eye/.kube/config
kind: Service
metadata:
  name: keycloak-postgresql-headless
  labels:
--
kind: Service
metadata:
  name: keycloak-postgresql-read
  labels:
--
kind: Service
metadata:
  name: keycloak-postgresql
  labels:
--
kind: Service
metadata:
  name: keycloak-headless
  labels:
--
kind: Service
metadata:
  name: keycloak-http
  labels:
/tmp/keycloak helm template keycloak . --set postgresql.replication.enabled=true --set postgresql.persistence.enabled=true | grep until -A 3
              until printf "." && nc -z -w 2 keycloak-postgresql-primary 5432; do
                  sleep 2;
              done;

```

So the service gets created without `-primary` suffix from the subchart, but the initContainer tries to ping `-primary` service which does not exist. Same goes for the Secret:

/tmp/keycloak helm template keycloak . --set postgresql.replication.enabled=true --set postgresql.persistence.enabled=true | grep secretkeyref -i -A 2
                secretKeyRef:
                  name: keycloak-postgresql
                  key: postgresql-replication-password
--
                secretKeyRef:
                  name: keycloak-postgresql
                  key: postgresql-postgres-password
--
                secretKeyRef:
                  name: keycloak-postgresql
                  key: postgresql-password
--
                secretKeyRef:
                  name: keycloak-postgresql
                  key: postgresql-replication-password
--
                secretKeyRef:
                  name: keycloak-postgresql
                  key: postgresql-postgres-password
--
                secretKeyRef:
                  name: keycloak-postgresql
                  key: postgresql-password
--
                secretKeyRef:
                  name: keycloak-postgresql-primary
                  key: postgresql-password
```

The usage of `postgresql.fullname` is wrong in this context, as the subchart uses bitnami's `common.names.fullname`  on those resources.

BR

**Comment by eyenx** [comment]
@vmaillot can you rebase?

**Comment by grieshaber** [comment]
Thanks for updating this PR. I will have a look and check the issue!

**Comment by vmaillot** [comment]
> Thanks for updating this PR. I will have a look and check the issue!

My pleasure, thanks man for your help, it's much appreciated!

**Review by grieshaber — APPROVED** [review]
Thanks for fixing and finding this!

**Comment by vmaillot** [comment]
> Thanks for fixing and finding this!

Awesome, no prob! As it breaks the CI, what's the plan? Could you force merge? Because it looks like I cannot do anything.

**Comment by grieshaber** [comment]
@vmaillot i'll check what i can do regarding the CI stuff; i don't want to break the ci in general, so some sort of fix is needed. How it'll look.. i'm not sure yet. I do not understand why it breaks exactly tbh, but will investigate now.

File: charts/keycloak/templates/_helpers.tpl [hunk 1]
```diff
@@ -65,7 +65,7 @@ Create a default fully qualified app name for the postgres requirement.
 */}}
 {{- define "keycloak.postgresql.fullname" -}}
 {{- $postgresContext := dict "Values" .Values.postgresql "Release" .Release "Chart" (dict "Name" "postgresql") -}}
-{{ include "postgresql.primary.fullname" $postgresContext }}
+{{ include "keycloak.name" .}}-{{ include "postgresql.name" $postgresContext }}
```
grieshaber, line 68:
```suggestion
{{ include "keycloak.fullname" .}}-{{ include "postgresql.name" $postgresContext }}
```

grieshaber (changes requested):
@vmaillot i got the root cause of the issue. This is not only a CI issue i think; we need to use `fullname` instead of `name` here, so that Release-Name etc. is considered correctly.

grieshaber (approved):

PR #490, opened 2021-10-02
[keycloak] feat: support helm chart version 2

### LINKED ISSUE #465: Error: apiVersion 'v2' is not valid. The value must be "v1"
Hey @unguiculus @thomasdarimont,

It seems that this chart is not compatible when using helm push to upload it.

```
CHART_NAME: keycloak
CHART_SEMVER: 14.0.1
"services" already exists with the same configuration, skipping
Error: error unpacking postgresql in keycloak: apiVersion 'v2' is not valid. The value must be "v1"
```

This comes because https://github.com/codecentric/helm-charts/blob/master/charts/keycloak/Chart.yaml#L1 is apiV1 and the postgres dependency is already apiV2 .

## Solution

* Bump keycloak helm chart apiVersion to v2

solidnerd:
It fixes the problem of https://github.com/codecentric/helm-charts/blob/master/charts/keycloak/Chart.yaml#L1 is apiV1 and the Postgres dependency is already apiV2.

Closes #465

### DIFF
--- a/charts/keycloak/Chart.lock
+++ b/charts/keycloak/Chart.lock
@@ -3,4 +3,4 @@ dependencies:
   repository: https://charts.bitnami.com/bitnami
   version: 10.3.13
 digest: sha256:0b35e861ab4e49a0d63eb83bf74a753a4b12d576d2d5f941701a659b4c93e1e4
-generated: "2021-03-18T11:41:44.099229802+01:00"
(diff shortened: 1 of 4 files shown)


solidnerd:
@mircohaug  Can you check it ?

mircohaug (approved):
LGTM. Sorry for the long wait

Issue #496, opened 2021-10-26
Ingress v1 : capabilities are not evaluated correctly


olevitt:
Hi,

as discussed in this commit : https://github.com/codecentric/helm-charts/commit/cb624cf356851ef3fe2cd0fba0762b1d485e44ff  
the check that is used on both mailhog and keycloak charts to figure out which ingress version to use (Kubernetes `1.22` which released like two months ago removed the old ingresses apiVersion so many users are now upgrading and need to have the new ingress class).  
From my testing, the current check `{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" -}}` (see https://github.com/codecentric/helm-charts/blob/master/charts/keycloak/templates/_helpers.tpl#L82) does not work when using `helm template` (or using workflows that rely on helm template such as `argoCD`.  
We could use the same type of check as in other charts (it seems most of the charts maintainers are in the process of migrating their ingress apiVersion check). For example https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/_helpers.tpl#L135
or
https://github.com/hashicorp/vault-helm/blob/main/templates/server-ingress.yaml#L13  

Any idea on the best check to implement ?

SeWieland:
Hmmm.. I do have the same issue, however in other charts the capability check works flawlessly!
I don't think a "manual value setting" is the resolution here...

E.g. the ingress of this chart works for me just fine: https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/templates/ingress.yaml

=> They have defined their capability check in the actual ingress.yaml and not in the helper.
But I have absoultely no cue why it won't work using the helper..

juliohm1978:
@SeWieland 
What Kubernetes version are you using? Up until 1.21, `networking.k8s.io/v1beta1` is still available, so the chart works when it sends the Ingress to that API.

On a side note: When you query the apiserver, it might be showing `networking.k8s.io/v1`, but that does not represent the actual api used to create the object. The apiserver prints an apiVersion it considers appropriate, but the object might've been added through an older endpoint (https://github.com/kubernetes/kubernetes/issues/58131).

SeWieland:
@juliohm1978 I'm on 1.21 and the chart still works. However the helm liniter produces the warning that the API will be deprecated in 1.22+. As this is the only warning I get in a lot of charts, I think this should be changed.


I already created a PR, however for both charts. As this is not accepted, I'll change that

PR #499, opened 2021-11-01
[keycloak] Allow ingress apiVersion to be specified directly in values.yaml instead of using Capabilities.

juliohm1978:
Fix https://github.com/codecentric/helm-charts/issues/496

Signed-off-by: juliohm1978 <jhm@juliohm.com.br>

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: keycloak
-version: 15.1.0
+version: 15.1.1
 appVersion: 15.0.2
 description: Open Source Identity and Access Management For Modern Applications and Services
 keywords:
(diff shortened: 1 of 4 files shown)


lesmyrmidons:
+1

olevitt:
Shouldn't the check here : https://github.com/codecentric/helm-charts/blob/f7914e5bb63ad63264fe024cbf735b5faf7d9320/charts/keycloak/templates/ingress.yaml#L95

be adapted to use the value overriden instead of re-checking using capabilities ? The apiVersion used need to be consistent otherwise we risk having an invalid syntax

SeWieland:
Please have a look at #501 / #502 as an alternative

juliohm1978:
Thanks! I think either one would work. But I confess there is an argument for usability, not requiring chart users to be aware of api versions used under the hood.

I'm all for approving #501 and #502, if that's the more popular choice :)

mircohaug:
Hi thanks for the PR but I prefer the usability of https://github.com/codecentric/helm-charts/pull/501

PR #501, opened 2021-11-10
[keycloak] Change Ingress Api Version Logic

### LINKED ISSUE #496: Ingress v1 : capabilities are not evaluated correctly

SeWieland:
Fix #496

Alternative to #499 => do not make the api version configurable, but retrieve it correctly.

I changed the logic for both charts => not yet fully tested!

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: keycloak
-version: 16.0.0
+version: 16.0.1
 appVersion: 15.0.2
 description: Open Source Identity and Access Management For Modern Applications and Services
 keywords:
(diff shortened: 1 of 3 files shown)


SeWieland:
Tested changes:
* Ingress uses `apiVersion: networking.k8s.io/v1` now at my setup
* Helm lint works without a warning

lesmyrmidons:
+1

mircohaug (dismissed):
LGTM

mircohaug (approved):

PR #506, opened 2021-11-22
[KeyCloak] Fix default service.httpManagementPort in README

kindlich:
Probably a copy-paste issue ^^

Default value taken from values.yaml
<https://github.com/codecentric/helm-charts/blob/6f9f62a5ad7ddffa12ea70d6c52ffe2512bfc79d/charts/keycloak/values.yaml#L249>


mircohaug:
Thanks for the fix. This was cherrypicked into e77407ca

PR #507, opened 2021-11-22
[keycloak] Add option to skip init containers

funkypenguin:
Hi guys,

This PR adds a non-default option to skip any initContainers when deploying the keycloak statefulset. 

Why?

Because when using a service mesh whose connectivity relies on a sidecar (like Istio's envoy-proxy), an initContainer which requires network connectivity will never complete.

Cheers!
D


mircohaug (approved):
LGTM

mircohaug (approved):

Issue #509, opened 2021-11-29
[Keycloack] 502 Bad Gateway - Proxy Buffer Size


c00:
A common error that happens with Keycloak on K8s is the 502 Bad Gateway error. \this has to do with the Proxy Buffer size of the Ingress. For example, [here's a solution](https://stackoverflow.com/a/68811151/1686932).

Since this is such a common occurence and the readme of this chart is super helpful in a variety of other ways, I want to suggest adding a note to the Readme to address this issue.

nyogendran:
I pretty much tried out all the options to fix the 502 Bad Gateway issue. But I could not.
Referred [Keycloak X](https://github.com/codecentric/helm-charts/blob/master/charts/keycloakx/values.yaml) to install Keycloak using Helm.

helm version
version.BuildInfo{Version:"v3.17.2", GitCommit:"cc0bbbd6d6276b83880042c1ecb34087e84d41eb", GitTreeState:"clean", GoVersion:"go1.24.1"}

% helm search repo codecentric/keycloak
NAME                    CHART VERSION   APP VERSION     DESCRIPTION                                       
codecentric/keycloak    18.10.0         17.0.1-legacy   Open Source Identity and Access Management For ...
codecentric/keycloakx   7.0.1           26.0.7          Keycloak.X - Open Source Identity and Access Ma...

% kubectl version  
Client Version: v1.32.2
Kustomize Version: v5.5.0
Server Version: v1.32.0

values-x.yaml
`fullnameOverride: ""

nameOverride: ""

replicas: 1

image:
  repository: "quay.io/keycloak/keycloak"
  tag: "26.0.7"
  pullPolicy: IfNotPresent

enableServiceLinks: true

updateStrategy: RollingUpdate

podManagementPolicy: Parallel

restartPolicy: Always

service:
  annotations: {}
  labels: {}
  type: LoadBalancer
  loadBalancerIP: ""
  httpPort: 8080
  httpNodePort: null
  httpsPort: 8443
  httpsNodePort: null
  extraPorts: []
  externalTrafficPolicy: "Cluster"
  sessionAffinity: ""
  sessionAffinityConfig: {}

serviceHeadless:
  enabled: true
  annotations: {}

skipInitContainers: false

terminationGracePeriodSeconds: 30

clusterDomain: cluster.local

command:
  - "/opt/keycloak/bin/kc.sh"
  - "start"
  - "--hostname-strict=false"

extraEnv: |
[...]

PR #510, opened 2021-11-29
Update README.md

c00:
Added a description on how to deal with the common "Bad Gateway" issue when using `keycloak` and `ingress-nginx` (with Proxy Protocol enabled)

The issue referring to this is #509


mircohaug:
Thanks for the fix. This was cherrypicked into 9076322f

PR #521, opened 2022-01-14
Fix labels rendering

Freaky-namuH:
Ingress labels currently appear under namespace and not label

Thanks for wanting to contribute.

Manual updates to the chart version are not needed any more. The version bumps are now based on commit messages. If you want to bump the major version include `major` in the commit message. For a feature release, include `feature` or `feat`. If you don't want to create a new release at all include `chore` in your commit message. The default is a new patch release. For the specific keywords have a look at [the script](scripts/bump-version.py).

Signed-off-by: Ashley Cambrell <acambrell@gmail.com>

### DIFF
--- a/charts/mailhog/templates/ingress.yaml
+++ b/charts/mailhog/templates/ingress.yaml
@@ -14,9 +14,9 @@ apiVersion: {{ $apiVersion }}
 kind: Ingress
 metadata:
   name: {{ $fullName }}
-  labels:
-    {{- include "mailhog.labels" . | nindent 4 }}
   namespace: {{ .Release.Namespace }}
+  labels:
+  {{- include "mailhog.labels" . | nindent 4 }}
   {{- with .Values.ingress.labels }}
     {{- toYaml . | nindent 4 }}
   {{- end }}


mircohaug:
I don't understand your change. Keys in a yaml document do not have a "order". Running `helm template --set ingress.enabled=true .` in the mailhog directory do produce  the following which from my point of view is correct. 

```yaml
---
# ... [SNIP] ...
# Source: mailhog/templates/ingress.yaml
kind: Ingress
metadata:
  name: RELEASE-NAME-mailhog
  labels:
    helm.sh/chart: mailhog-5.0.3
    app.kubernetes.io/name: mailhog
    app.kubernetes.io/instance: RELEASE-NAME
    app.kubernetes.io/version: "v1.0.1"
  namespace: infra
spec:
# ... [SNIP] ...
```

If you have further information why this change is necessary feel free to reopen the PR.

Freaky-namuH:
The trouble is that there are labels above and below the namespace key. When you add ingress labels which we do for cost attribution purposes, the ingress labels come AFTER the namespace which creates invalid YAML syntax.

Mailhog Label Fix

Issue #309, opened 2020-09-28
[Feature] Extra Prometheus metrics


ath88:
For the Keycloak 11 rewrite of the chart, it seems that the supported metrics were changed away from the Keycloak Metrics SPI[1], to use the internal Wildfly Prometheus metrics[2]. They provide quite different metrics, as far as I can see. The SPI exposes keycloak events, such as signup and logins.

If I wanted to add the SPI, I could just add it as an init container. But, I would need to add an additional ServiceMonitor manually, since the chart does not support the additional configuration.

How about the chart support the SPI, or at least multiple scrape endpoints?

[1] https://github.com/aerogear/keycloak-metrics-spi
[2] https://docs.wildfly.org/16/Admin_Guide.html#exposed-metrics

Tim-Schwalbe:
Also the metrics endpoint is not secured with a password if I just add the SPI. The metrics endpoint is just reachable at `/auth/realms/master/metrics`

Any Idea how to fix this?

ath88:
There is no solution from the developers - but it should be trivial to configure nginx to block it, or to add authentication.

See https://github.com/aerogear/keycloak-metrics-spi/issues/39

unguiculus:
You can add an init container or use a custom image. Adding an option to support multiple scrape endpoints is certainly something I can look into.

ath88:
I'll gladly add it as an init container - thats what I did before. But having to manage an extra servicemonitor, along with the helm chart, is cumbersome. :)

Tim-Schwalbe:
Yes I did that and also needed to lock the endpoint for the metrics.

Here is my NGINX Ingress Config to block requests to /auth/realms/master/metrics. I did not test if there will be an endpoint for every realm, but for now it was working.

```
   annotations: 
    kubernetes.io/ingress.class: "nginx"
    kubernetes.io/tls-acme: "true"
    nginx.ingress.kubernetes.io/server-snippet: |
      location /auth/realms/master/metrics {
          return 403;
      }
```

unguiculus:
@ath88 Please checkout #321. Does this suit your needs?

ath88:
It does indeed look like it would suit my needs! Thanks a lot! :)

harshitmahapatra:
@Tim-Schwalbe Would you mind sharing your init container config?

Tim-Schwalbe:
Here it is:

```
extraInitContainers: |
  - name: extensions
    image: busybox
[...]

PR #533, opened 2022-01-28
[keycloakx] KeycloakX support [DRAFT]

thomasdarimont:
A revised chart for Keycloak.X

### DIFF
--- a/charts/keycloakx/README.md
+++ b/charts/keycloakx/README.md
@@ -1,0 +1,574 @@
+# Keycloak-X
+
+[Keycloak-X](http://www.keycloak.org/) is an open source identity and access management for modern applications and services.
+
+Note that this chart is the logical successor of the Wildfly based [codecentric/keycloak](https://github.com/codecentric/helm-charts/tree/master/charts/keycloak) chart.
+
+## TL;DR;
+
+```console
+$ helm install keycloak codecentric/keycloakx
+```
+
+## Introduction
+
+This chart bootstraps a [Keycloak](http://www.keycloak.org/) StatefulSet on a [Kubernetes](https://kubernetes.io) cluster using the [Helm](https://helm.sh) package manager.
+It provisions a fully featured Keycloak installation.
+For more information on Keycloak and its capabilities, see its [documentation](http://www.keycloak.org/documentation.html).
+
+## Installing the Chart
+
+To install the chart with the release name `keycloakx`:
+
+```console
+$ helm install keycloak codecentric/keycloakx
+```
+
+## Uninstalling the Chart
+
+To uninstall the `keycloakx` deployment:
+
+```console
+$ helm uninstall keycloakx
+```
+
+## Configuration
+
+The following table lists the configurable parameters of the Keycloak-X chart and their default values.
+
+| Parameter                                    | Description                                                                                                                                                                                                                                                                       | Default                                                                                                                               |
+|----------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------|
+| `fullnameOverride`                           | Optionally override the fully qualified name                                                                                                                                                                                                                                      | `""`                                                                                                                                  |
+| `nameOverride`                               | Optionally override the name                                                                                                                                                                                                                                                      | `""`                                                                                                                                  |
+| `replicas`                                   | The number of replicas to create                                                                                                                                                                                                                                                  | `1`                                                                                                                                   |
+| `image.repository`                           | The Keycloak image repository                                                                                                                                                                                                                                                     | `quay.io/keycloak/keycloak`                                                                                                           |
+| `image.tag`                                  | Overrides the Keycloak image tag whose default is the chart version                                                                                                                                                                                                               | `""`                                                                                                                                  |
+| `image.pullPolicy`                           | The Keycloak image pull policy                                                                                                                                                                                                                                                    | `IfNotPresent`                                                                                                                        |
+| `imagePullSecrets`                           | Image pull secrets for the Pod                                                                                                                                                                                                                                                    | `[]`                                                                                                                                  |
+| `hostAliases`                                | Mapping between IPs and hostnames that will be injected as entries in the Pod's hosts files                                                                                                                                                                                       | `[]`                                                                                                                                  |
+| `enableServiceLinks`                         | Indicates whether information about services should be injected into Pod's environment variables, matching the syntax of Docker links                                                                                                                                             | `true`                                                                                                                                |
+| `podManagementPolicy`                        | Pod management policy. One of `Parallel` or `OrderedReady`                                                                                                                                                                                                                        | `Parallel`                                                                                                                            |
+| `restartPolicy`                              | Pod restart policy. One of `Always`, `OnFailure`, or `Never`                                                                                                                                                                                                                      | `Always`                                                                                                                              |
+| `serviceAccount.create`                      | Specifies whether a ServiceAccount should be created                                                                                                                                                                                                                              | `true`                                                                                                                                |
+| `serviceAccount.name`                        | The name of the service account to use. If not set and create is true, a name is generated using the fullname template                                                                                                                                                            | `""`                                                                                                                                  |
+| `serviceAccount.annotations`                 | Additional annotations for the ServiceAccount                                                                                                                                                                                                                                     | `{}`                                                                                                                                  |
+| `serviceAccount.labels`                      | Additional labels for the ServiceAccount                                                                                                                                                                                                                                          | `{}`                                                                                                                                  |
+| `serviceAccount.imagePullSecrets`            | Image pull secrets that are attached to the ServiceAccount                                                                                                                                                                                                                        | `[]`                                                                                                                                  |
+| `rbac.create`                                | Specifies whether RBAC resources are to be created                                                                                                                                                                                                                                | `false`
+| `rbac.rules`                                 | Custom RBAC rules, e. g. for KUBE_PING                                                                                                                                                                                                                                            | `[]`
+| `podSecurityContext`                         | SecurityContext for the entire Pod. Every container running in the Pod will inherit this SecurityContext. This might be relevant when other components of the environment inject additional containers into running Pods (service meshes are the most prominent example for this) | `{"fsGroup":1000}`                                                                                                                    |
+| `securityContext`                            | SecurityContext for the Keycloak container                                                                                                                                                                                                                                        | `{"runAsNonRoot":true,"runAsUser":1000}`                                                                                              |
+| `extraInitContainers`                        | Additional init containers, e. g. for providing custom themes                                                                                                                                                                                                                     | `[]`                                                                                                                                  |
+| `skipInitContainers`                         | Skip all init containers (to avoid issues with service meshes which require sidecar proxies for connectivity)                                                                                                                                                                     | `false`
+| `extraContainers`                            | Additional sidecar containers, e. g. for a database proxy, such as Google's cloudsql-proxy                                                                                                                                                                                        | `[]`                                                                                                                                  |
+| `lifecycleHooks`                             | Lifecycle hooks for the Keycloak container                                                                                                                                                                                                                                        | `{}`                                                                                                                                  |
+| `terminationGracePeriodSeconds`              | Termination grace period in seconds for Keycloak shutdown. Clusters with a large cache might need to extend this to give Infinispan more time to rebalance                                                                                                                        | `60`                                                                                                                                  |
+| `clusterDomain`                              | The internal Kubernetes cluster domain                                                                                                                                                                                                                                            | `cluster.local`                                                                                                                       |
+| `command`                                    | Overrides the default entrypoint of the Keycloak container                                                                                                                                                                                                                        | `[]`                                                                                                                                  |
+| `args`                                       | Overrides the default args for the Keycloak container                                                                                                                                                                                                                             | `[]`                                                                                                                                  |
+| `extraEnv`                                   | Additional environment variables for Keycloak                                                                                                                                                                                                                                     | `""`                                                                                                                                  |
+| `extraEnvFrom`                               | Additional environment variables for Keycloak mapped from a Secret or ConfigMap                                                                                                                                                                                                   | `""`                                                                                                                                  |
+| `priorityClassName`                          | Pod priority class name                                                                                                                                                                                                                                                           | `""`                                                                                                                                  |
+| `affinity`                                   | Pod affinity                                                                                                                                                                                                                                                                      | Hard node and soft zone anti-affinity                                                                                                 |
+| `topologySpreadConstraints`                  | Topology spread constraints                                                                                                                                                                                                                                                       | Constraints used to spread pods                                                                                                       |
+| `nodeSelector`                               | Node labels for Pod assignment                                                                                                                                                                                                                                                    | `{}`                                                                                                                                  |
+| `tolerations`                                | Node taints to tolerate                                                                                                                                                                                                                                                           | `[]`                                                                                                                                  |
+| `podLabels`                                  | Additional Pod labels                                                                                                                                                                                                                                                             | `{}`                                                                                                                                  |
+| `podAnnotations`                             | Additional Pod annotations                                                                                                                                                                                                                                                        | `{}`                                                                                                                                  |
+| `livenessProbe`                              | Liveness probe configuration                                                                                                                                                                                                                                                      | `{"httpGet":{"path":"/auth/","port":"http"},"initialDelaySeconds":0,"timeoutSeconds":5}`                                              |
+| `readinessProbe`                             | Readiness probe configuration                                                                                                                                                                                                                                                     | `{"httpGet":{"path":"/auth/realms/master","port":"http"},"initialDelaySeconds":30,"timeoutSeconds":1}`                                |
+| `startupProbe`                               | Startup probe configuration                                                                                                                                                                                                                                                       | `{"httpGet":{"path":"/auth/","port":"http"},"initialDelaySeconds":30,"timeoutSeconds":5, "failureThreshold": 60, "periodSeconds": 5}` |
+| `resources`                                  | Pod resource requests and limits                                                                                                                                                                                                                                                  | `{}`                                                                                                                                  |
+| `extraVolumes`                               | Add additional volumes, e. g. for custom themes                                                                                                                                                                                                                                   | `""`                                                                                                                                  |
+| `extraVolumeMounts`                          | Add additional volumes mounts, e. g. for custom themes                                                                                                                                                                                                                            | `""`                                                                                                                                  |
+| `extraPorts`                                 | Add additional ports, e. g. for admin console or exposing JGroups ports                                                                                                                                                                                                           | `[]`                                                                                                                                  |
+| `podDisruptionBudget`                        | Pod disruption budget                                                                                                                                                                                                                                                             | `{}`                                                                                                                                  |
+| `statefulsetAnnotations`                     | Annotations for the StatefulSet                                                                                                                                                                                                                                                   | `{}`                                                                                                                                  |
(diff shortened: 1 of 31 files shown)


Morl99:
Wanted to leave some positive feedback, we now use this version of the chart for deployment. Thanks!

charts/keycloakx/templates/ingress.yaml
```diff
@@ -0,0 +1,120 @@
...
+    {{- range $ingress.tls }}
+    - hosts:
+      {{- range .hosts }}
+        - {{ tpl . $ | quote }}
+      {{- end }}
+      {{- with .secretName }}
+      secretName: {{ tpl . $ }}
+      {{- end }}
+    {{- end }}
+{{- end }}
```
vanrar68, line 97: Would be nice to have a distinct tls configuration for console (see #549)

charts/keycloakx/templates/_helpers.tpl
```diff
@@ -0,0 +1,76 @@
...
+{{- .Values.fullnameOverride | trunc 20 | trimSuffix "-" }}
+{{- else }}
+{{- $name := default .Chart.Name .Values.nameOverride }}
+{{- if contains $name .Release.Name }}
+{{- .Release.Name | trunc 20 | trimSuffix "-" }}
+{{- else }}
+{{- printf "%s-%s" .Release.Name $name | trunc 20 | trimSuffix "-" }}
+{{- end }}
+{{- end }}
+{{- end }}
```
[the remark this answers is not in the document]
thomasdarimont: Thanks for the feedback! I'll adjust the code to truncate the name with 63 characters

charts/keycloakx/templates/ingress.yaml
thomasdarimont: Thanks for the feedback, I'll give this a try.

charts/keycloakx/values.yaml
```diff
@@ -0,0 +1,494 @@
...
+hostAliases: []
+# - ip: "1.2.3.4"
+#   hostnames:
+#     - "my.host.com"
+
+# Indicates whether information about services should be injected into Pod's environment variables, matching the syntax of Docker links
+enableServiceLinks: true
+
+# Pod management policy. One of `Parallel` or `OrderedReady`
+podManagementPolicy: Parallel
```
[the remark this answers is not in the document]
thomasdarimont: This makes sense! Thanks again

charts/keycloakx/values.yaml
```diff
@@ -0,0 +1,494 @@
...
+    port: http
+  initialDelaySeconds: 0
+  timeoutSeconds: 5
+
+# Readiness probe configuration
+readinessProbe: |
+  httpGet:
+    path: /auth/realms/master
+    port: http
+  initialDelaySeconds: 30
```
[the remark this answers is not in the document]
thomasdarimont: I'd set this to 10 seconds as some custom keycloakx image might use `--auto-build` to dynamically register extensions.
If you want this to start faster you can override the `initialDelaySeconds`

charts/keycloakx/values.yaml
```diff
@@ -0,0 +1,494 @@
...
+  timeoutSeconds: 1
+
+# Startup probe configuration
+startupProbe: |
+  httpGet:
+    path: /auth/
+    port: http
+  initialDelaySeconds: 30
```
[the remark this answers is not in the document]
thomasdarimont: I agree

charts/keycloakx/values.yaml
```diff
@@ -0,0 +1,494 @@
...
+  # The HTTP Service node port if type is NodePort
+  httpNodePort: null
+  # The HTTPS Service port
+  httpsPort: 8443
+  # The HTTPS Service node port if type is NodePort
+  httpsNodePort: null
+  # The WildFly management Service port
+  httpManagementPort: 9990
+  # The WildFly management Service node port if type is NodePort
+  httpManagementNodePort: null
```
PSanetra, line 247: Here are also some mentions of wildfly.

charts/keycloakx/README.md
```diff
@@ -0,0 +1,628 @@
...
+It must be configured via environment variables:
+
+* `KC_DB_USERNAME` or `KC_DB_USERNAME_FILE`
+* `KC_DB_PASSWORD` or `KC_DB_PASSWORD_FILE`
+
+Please refer to the section on database configuration for how to configure a secret for this.
+
+### High Availability and Clustering
+
+TODO Revise
```
PSanetra, line 335: Here is still a todo. Should this be resolved before merging?

PSanetra, line 335: Here is still a todo. Should this be resolved before merging?

charts/keycloakx/README.md
```diff
@@ -0,0 +1,628 @@
...
+A common issue with Keycloak and nginx is that the proxy buffer may be too small for what Keycloak is trying to send. This will result in a Bad Gateway (502) error. There are [many](https://github.com/kubernetes/ingress-nginx/issues/4637) [issues](https://stackoverflow.com/questions/56126864/why-do-i-get-502-when-trying-to-authenticate) around the internet about this. The solution is to increase the buffer size of nginx. This can be done by creating an annotation in the ingress specification:
+
+```yaml
+ingress:
+  annotations:
+    nginx.ingress.kubernetes.io/proxy-buffer-size: "128k"
+```
+
+## Upgrading
+
```
PSanetra, line 628: Upgrading section is still empty

PSanetra, line 628: Upgrading section is still empty

charts/keycloakx/README.md
```diff
@@ -0,0 +1,628 @@
...
+It provisions a fully featured Keycloak installation.
+For more information on Keycloak and its capabilities, see its [documentation](http://www.keycloak.org/documentation.html).
+
+## Prerequisites Details
+
+The chart has an optional dependency on the [PostgreSQL](https://github.com/bitnami/charts/tree/master/bitnami/postgresql) chart.
+By default, the PostgreSQL chart requires PV support on underlying infrastructure (may be disabled).
+
+**Important**
+Since Version `v1.0.0` the support for Helm v2 is dropped in favor of to support also the latest postgres Helm Chart Version which requires Helm v3 only.
```
PSanetra, line 23: This paragraph is probably not needed in this chart as it did never support Helm v2, right?

PSanetra, line 23: This paragraph is probably not needed in this chart as it did never support Helm v2, right?

charts/keycloakx/templates/statefulset.yaml
```diff
@@ -0,0 +1,192 @@
...
+            {{- end }}
+          ports:
+            - name: http
+              containerPort: 8080
+              protocol: TCP
+            - name: https
+              containerPort: 8443
+              protocol: TCP
+            - name: http-management
+              containerPort: 9990
```
PSanetra, line 119: Is the declaration of the http-management port still needed? Might also be removed from the service.

PSanetra, line 119: Is the declaration of the http-management port still needed? Might also be removed from the service.

charts/keycloakx/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,46 @@
...
+        - podSelector:
+            matchLabels:
+              {{- include "keycloak.selectorLabels" . | nindent 14 }}
+      ports:
+        - protocol: TCP
+          port: 8080
+        - protocol: TCP
+          port: 8443
+        - protocol: TCP
+          port: 9990
```
PSanetra, line 41: Port 9990 might be removed here, too.

PSanetra, line 41: Port 9990 might be removed here, too.

charts/keycloakx/templates/ingress.yaml
```diff
@@ -0,0 +1,131 @@
...
+{{- if $ingress.enabled -}}
+{{- $apiV1 := false -}}
+{{- $apiVersion := "extensions/v1beta1" -}}
+{{- if and (.Capabilities.APIVersions.Has "networking.k8s.io/v1") (semverCompare ">= v1.19.0-0" .Capabilities.KubeVersion.Version) -}}
+  {{- $apiVersion = "networking.k8s.io/v1" -}}
+  {{- $apiV1 = true -}}
+{{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" -}}
+  {{- $apiVersion = "networking.k8s.io/v1beta1" -}}
+{{- else -}}
+  {{- $apiVersion = "extensions/v1beta1" -}}
```
PSanetra, line 11: Do we plan to drop support for those old api version at some point in time? This might simplify the ingress resource templates.

PSanetra, line 11: Do we plan to drop support for those old api version at some point in time? This might simplify the ingress resource templates.

charts/keycloakx/templates/configmap-startup.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.startupScripts }}
+{{- $highAvailability := gt (int .Values.replicas) 1 -}}
```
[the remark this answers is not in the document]
thomasdarimont: Thanks! I removed the variable

charts/keycloakx/templates/service-headless.yaml
```diff
@@ -0,0 +1,18 @@
+{{- $highAvailability := gt (int .Values.replicas) 1 -}}
```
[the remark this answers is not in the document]
thomasdarimont: Thanks! I removed the variable

charts/keycloakx/values.yaml
thomasdarimont: Good catch!

charts/keycloakx/values.yaml
```diff
@@ -0,0 +1,494 @@
...
+  # Additional labels for the ServiceMonitor
+  labels: {}
+  # Interval at which Prometheus scrapes metrics
+  interval: 10s
+  # Timeout for scraping
+  scrapeTimeout: 10s
+  # The path at which metrics are served
+  path: /metrics
+  # The Service port at which metrics are served
+  port: http-management
```
PSanetra, line 398: The "non-extra" service monitor references the http-management port on the metrics endpoint by default. Does this still make sense?

charts/keycloakx/values.yaml
```diff
@@ -0,0 +1,494 @@
...
+
+## Overrides the default args for the Keycloak container
+args: []
+
+# Additional environment variables for Keycloak
+extraEnv: ""
+  # - name: KEYCLOAK_LOGLEVEL
+  #   value: DEBUG
+  # - name: WILDFLY_LOGLEVEL
+  #   value: DEBUG
```
PSanetra, line 109: Here is still wildfly mentioned in the example

charts/keycloakx/values.yaml
```diff
@@ -0,0 +1,494 @@
...
+        podAffinityTerm:
+          labelSelector:
+            matchLabels:
+              {{- include "keycloak.selectorLabels" . | nindent 12 }}
+            matchExpressions:
+              - key: app.kubernetes.io/component
+                operator: NotIn
+                values:
+                  - test
+          topologyKey: failure-domain.beta.kubernetes.io/zone
```
PSanetra, line 145: This label is deprecated as of Kubernetes v1.17: https://kubernetes.io/docs/reference/labels-annotations-taints/#failure-domainbetakubernetesiozone

charts/keycloakx/templates/ingress.yaml
thomasdarimont: Good point, which api versions should we keep?

charts/keycloakx/values.yaml
```diff
@@ -0,0 +1,494 @@
...
+    ## See https://www.ibm.com/support/pages/502-error-ingress-keycloak-response
+    # nginx.ingress.kubernetes.io/proxy-buffer-size: 128k
+
+  # Additional Ingress labels
+  labels: {}
+   # List of rules for the Ingress
+  rules:
+    -
+      # Ingress host
+      host: '{{ .Release.Name }}.keycloak.example.com'
```
PSanetra, line 283: Does it make sense to declare this host as default value? I would rather provide this just as an example comment, no? (Same in tls section)

charts/keycloakx/values.yaml
```diff
@@ -0,0 +1,494 @@
...
+    enabled: false
+    # The name of Ingress Class associated with the console ingress only
+    ingressClassName: ""
+    # Ingress annotations for console ingress only
+    # Useful to set nginx.ingress.kubernetes.io/whitelist-source-range particularly
+    annotations: {}
+    rules:
+      -
+        # Ingress host
+        host: '{{ .Release.Name }}.keycloak.example.com'
```
PSanetra, line 306: Same here regarding the example comment

charts/keycloakx/values.yaml
```diff
@@ -0,0 +1,494 @@
...
+      cpu: "20m"
+      memory: "32Mi"
+
+postgresql:
+  # If `true`, the Postgresql dependency is enabled
+  enabled: true
+  # PostgreSQL User to create
+  postgresqlUsername: keycloak
+  # PostgreSQL Password for the new user
+  postgresqlPassword: keycloak
```
PSanetra, line 373: I would rather set no default password here and require the user to set a password explicitly.

charts/keycloakx/values.yaml
thomasdarimont: Removed those variables

charts/keycloakx/README.md
thomasdarimont: Revised the HA section

charts/keycloakx/README.md
thomasdarimont: Added note

charts/keycloakx/README.md
thomasdarimont: Removed the paragraph

charts/keycloakx/templates/statefulset.yaml
thomasdarimont: Removed

charts/keycloakx/templates/networkpolicy.yaml
thomasdarimont: Removed

charts/keycloakx/templates/service-http.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - name: https
+      port: {{ .Values.service.httpsPort }}
+      targetPort: https
+      {{- if and (or (eq "NodePort" .Values.service.type) (eq "LoadBalancer" .Values.service.type) ) .Values.service.httpsNodePort }}
+      nodePort: {{ .Values.service.httpsNodePort }}
+      {{- end }}
+      protocol: TCP
+    - name: http-management
+      port: {{ .Values.service.httpManagementPort }}
+      targetPort: http-management
```
PSanetra, line 53: service is still targeting the removed http-management port.

charts/keycloakx/values.yaml
thomasdarimont: Metrics will be served via the /metrics endpoint via the http/https port if `KC_METRICS_ENABLED=true`.
https://www.keycloak.org/server/all-config?q=metric

thomasdarimont: I changed this to http

charts/keycloakx/values.yaml
thomasdarimont: Removed those

charts/keycloakx/values.yaml
thomasdarimont: Changed to `topology.kubernetes.io/zone`

charts/keycloakx/values.yaml
thomasdarimont: Commented out

charts/keycloakx/values.yaml
thomasdarimont: Commented

charts/keycloakx/values.yaml
thomasdarimont: Commented

charts/keycloakx/templates/ingress.yaml
PSanetra: v1 was stable with Kubernetes v1.19.0, which was released on Aug 26, 2020. So I guess that should have been enough time to upgrade. I would suggest to keep v1 and maybe add other versions again if someone requests it via an issue.

charts/keycloakx/templates/service-http.yaml
thomasdarimont: Thanks, removed

charts/keycloakx/values.yaml
```diff
@@ -0,0 +1,484 @@
...
+  # postgresqlPassword: keycloak
+  # PostgreSQL Database to create
+  postgresqlDatabase: keycloak
+  # PostgreSQL network policy configuration
+  networkPolicy:
+    enabled: false
+
+serviceMonitor:
+  # If `true`, a ServiceMonitor resource for the prometheus-operator is created
+  enabled: false
```
PSanetra, line 372: Does it make sense to automatically set `KC_METRICS_ENABLED=true` if `serviceMonitor.enabled` is `true`?

charts/keycloakx/values.yaml
```diff
@@ -0,0 +1,484 @@
...
+  # Interval at which Prometheus scrapes metrics
+  interval: 10s
+  # Timeout for scraping
+  scrapeTimeout: 10s
+  # The path at which metrics are served
+  path: /metrics
+  # The Service port at which metrics are served
+  port: http
+
+extraServiceMonitor:
```
PSanetra, line 390: Does it make sense to remove or extend the `extraServiceMonitor` option? The common use case seems to be to monitor the `/auth/realms/master/metrics` endpoint, but would we not want to support more than one realm metrics endpoint? At the moment only a single additional endpoint is supported.

charts/keycloakx/templates/ingress.yaml
thomasdarimont: Thanks again! I just removed support for the version switch.

charts/keycloakx/values.yaml
thomasdarimont: Yes indeed, do you have a quick snippet for this?

charts/keycloakx/values.yaml
thomasdarimont: Well there are `/auth/metrics` for global metrics and `/auth/realms/foo/metrics` with potentially realm specific metrics.
Of course the latter depends on the metrics extension used.

I also just noticed that we should make the `/auth` path option (`--http-relative-path` or `KC_HTTP_RELATIVE_PATH`) configurable which also defines the prefix for `/metrics`.

charts/keycloakx/templates/statefulset.yaml
```diff
@@ -0,0 +1,189 @@
...
+            - name: KC_DB_DATABASE
+              value: {{ .Values.postgresql.postgresqlDatabase | quote }}
+            - name: KC_DB_USERNAME
+              value: {{ .Values.postgresql.postgresqlUsername | quote }}
+            - name: KC_DB_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "keycloak.postgresql.fullname" . }}
+                  key: postgresql-password
+            {{- end }}
```
PSanetra, line 103: I guess this addition would be nice when the serviceMonitor is enabled:
```suggestion
            {{- end }}
            {{- if .Values.serviceMonitor.enabled }}
            - name: KC_METRICS_ENABLED
              value: "true"
            {{- end }}
```

thomasdarimont: Great! Adding this

charts/keycloakx/values.yaml
PSanetra: ok, maybe we should also not change too much here to not break too much compatibility with the old Keycloak helm chart. Just as reference: The original issue requesting the `extraServiceMonitor` option seems to be this one: https://github.com/codecentric/helm-charts/issues/309

PR #537, opened 2022-02-02
[mailhog] Incorrect YAML syntax when Ingress labels are used

Freaky-namuH:
Signed-off-by: Ashley Cambrell <acambrell@gmail.com>


Freaky-namuH:
The template has labels above and below the namespace key. When you add ingress labels, which we do for cost attribution purposes, the ingress labels come AFTER the namespace which creates invalid YAML syntax.

Mailhog Label Fix

mircohaug (approved):
LGTM. Thanks for the super extensive problem description.

PR #538, opened 2022-02-02
[mailhog] fix: auth volume overwritten when outgoingSMTP.enabled

dnplkndll:
now:
when outgoingSMTP.enabled and auth.enabled the SMTP volume write to the root volumes key

fix:
make the volume key either or and add to the array of volume to support either or both

log:
```
➜ ~ k logs -f -n $n $p
2022/02/02 05:32:58 Using in-memory storage
[HTTP] Error reading auth-file: open /authdir/auth.txt: no such file or directory
```

### DIFF
--- a/charts/mailhog/templates/deployment.yaml
+++ b/charts/mailhog/templates/deployment.yaml
@@ -66,17 +66,18 @@ spec:
           readinessProbe:
             tcpSocket:
               port: tcp-smtp
-          {{- if .Values.auth.enabled }}
+          {{ if or .Values.auth.enabled .Values.outgoingSMTP.enabled }}
           volumeMounts:
+            {{- if .Values.auth.enabled }}
             - name: authdir
               mountPath: /authdir
               readOnly: true
-          {{- end }}
-          {{- if .Values.outgoingSMTP.enabled }}
-          volumeMounts:
+            {{- end }}
+            {{- if .Values.outgoingSMTP.enabled }}
             - name: outsmtpdir
               mountPath: /config
               readOnly: true
+            {{- end }}
           {{- end }}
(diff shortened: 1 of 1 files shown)


dnplkndll:
@thomasdarimont tested this and resolves the issue with both on at the same time, did not test every permutation. but hoping to get this merged so I do not have to publish from a fork.

dnplkndll:
@max3903 are you guys still using this email trap?

max3903:
@dnplkndll Yes

max3903:
Ping @JevinD

dnplkndll:
@patrys I saw you reference the image. do you use this chart or just the image? https://github.com/saleor/saleor-platform/blob/main/docker-compose.yml#L112

charts/mailhog/templates/deployment.yaml
```diff
@@ -96,15 +97,16 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      {{- if .Values.auth.enabled }}
+      {{ if or .Values.auth.enabled .Values.outgoingSMTP.enabled }}
       volumes:
+      {{ end }}
```
mircohaug, line 102: Please move this `{{ end }}` to the end of the volumes block. This works as well but the proposed change is more clear. I would have added it to the PR myself bu you did not allow PR edits...

mircohaug (changes requested):
Thanks for contributing. Please fix this tiny thing and I will merge the PR.

dnplkndll:
> Thanks for contributing. Please fix this tiny thing and I will merge the PR.

anything else you need on this?

mircohaug (approved):

Issue #548, opened 2022-02-23
Use of distinct TLS config for keycloak ingress and keycloak-console ingress


vanrar68:
It would be convenient to be able to configure distinct TLS configuration for keycloak ingress and keycloak-console ingress, especially if ingress fetches certificates from a cert issuer. 
Currently the chart forces keycloak-console ingress to use the same TLS config as keycloak ingress.

PR #549, opened 2022-02-23
Support different TLS config for keycloak ingress and keycloak-consol…

vanrar68:
Attempt to solve #548 
Make ingress.console use ingress.console.tls if available otherwise fallback to ingress.tls (if any)


thomasdarimont:
@vanrar68 thanks for this PR! I assume you want to use this to serve the admin-console with a different domain.
Would be great if you could document the option in the [readme](https://github.com/codecentric/helm-charts/blob/master/charts/keycloak/README.md)

vanrar68:
> @vanrar68 thanks for this PR! I assume you want to use this to serve the admin-console with a different domain. Would be great if you could document the option in the [readme](https://github.com/codecentric/helm-charts/blob/master/charts/keycloak/README.md)

@thomasdarimont done. I also updated the values.yaml with a sample config (but left the default ingress.console.enabled = false). Let me know if you'd rather leave values.yaml untouched

mircohaug:
Hey @vanrar68 Thanks for the contribution. I moved your sample config into comments in order to not break existing deployments with the console ingress enabled. In this case the new "sample" TLS config would have taken precedence over the old behavior. I also squashed the commits for a clearer git history in main. Take care

vanrar68:
@thomasdarimont any reason to not fully include these changes in the new KeycloakX chart ? 
KeycloakX chart misses the documentation part (README.md) and the default values part (values.yaml)

thomasdarimont:
Thanks for the hint. That was an oversight when we ported the keycloak chart to keycloakx.
I just added a PR for this #565

PR #566, opened 2022-04-09
Mailhog: Quote namespace

dpunkturban:
Signed-off-by: Dennis Urban <mail@dpunkturban.io>

Hi,

i quoted all namespaces in the mailhog chart, so it can be used in namespaces with numeric names. Without quoting, i get the following error because the numeric namespace is interpreted as number and not string:

```
❯ helm install mailhog .
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: unable to decode "": json: cannot unmarshal number into Go struct field ObjectMeta.metadata.namespace of type string
```


thomasdarimont (approved):
LGTM

thomasdarimont:
Thanks @dpunkturban for your contribution!

PR #573, opened 2022-04-12
feat: make test deletion policy configurable

jnodorp-jaconi:
Signed-off-by: Julian Nodorp <jnodorp@jaconi.io>

Background: When using a cluster autoscaler, unmanaged pods prevent the removal of nodes. This is true, even for completed pods (such as completed pods created by `helm test`). Setting the `helm.sh/hook-delete-policy` annotation to `hook-succeeded` solves this problem.

This PR makes the `helm.sh/hook-delete-policy` annotation configurable. The default value is `before-hook-creation` which is the implicit default already (see https://helm.sh/docs/topics/charts_hooks/#hook-deletion-policies).


PSanetra:
I did not know about this hook feature. Does it even make sense to default to `hook-succeeded` or can you think of common cases where it makes sense to keep this test pod if the test succeeded?

jnodorp-jaconi:
> I did not know about this hook feature. Does it even make sense to default to `hook-succeeded` or can you think of common cases where it makes sense to keep this test pod if the test succeeded?

Not really. I just did not want to change default behavior.

thomasdarimont:
Thanks @jnodorp-jaconi for your contribution!

PR #574, opened 2022-04-14
[keycloakx] use http.relativePath instead hardcoded '/auth', correct README

dro-sh:
Signed-off-by: Serfi7 <drozhzhin200@yandex.ru>

Support `http.relativePath` in places, where `/auth` was hardcoded.
Correct README in some places


thomasdarimont (commented):
LGTM any remarks @mircohaug  @PSanetra ?

charts/keycloakx/README.md
```diff
@@ -482,46 +482,18 @@ extraVolumes: |
...
+  relativePath: '/'
 ```
 
-The above YAML references introduces the custom value `contextPath` which is possible because `startupScripts`, `livenessProbe`, `readinessProbe`, and `startupProbe` are templated using the `tpl` function.
-Note that it must not start with a slash.
 Alternatively, you may supply it via CLI flag:
 
 ```console
---set-string contextPath=mycontext
+--set-string http.relativePath=/
```
thomasdarimont, line 496: This needs to be documented as a breaking change, e.g. in the upgrade guide section in the readme.

This also requires a minor version bump.

thomasdarimont:
@dro-sh thank you for your contribution, we'll take a look :)

To ease the automatic versioning, could you prefix your commit message with `feature:` ?

charts/keycloakx/README.md
PSanetra: It is only a breaking change when migrating from the old Keycloak chart, right? The `http.relativePath` is already available in master, but the readme was not adjusted in the KeycloakX chart.

charts/keycloakx/values.yaml
```diff
@@ -275,15 +275,15 @@ ingress:
       host: '{{ .Release.Name }}.keycloak.example.com'
       # Paths for the host
       paths:
-        - path: /
+        - path: '{{ trimSuffix "/" .Values.http.relativePath}}/'
```
PSanetra, line 278: I am not sure if this is correct. The ingress should forward all requests to the keycloak service, regardless of the fact that there might be a relativePath configured or not, right?

charts/keycloakx/values.yaml
```diff
@@ -275,15 +275,15 @@ ingress:
...
           pathType: Prefix
   # TLS configuration
   tls:
     - hosts:
         - keycloak.example.com
       secretName: ""
 
-  # ingress for console only (/auth/admin)
+  # ingress for console only ({{ trimSuffix "/" .http.relativePath}}/admin)
```
PSanetra, line 286: I think I would not include this in the comment.

charts/keycloakx/values.yaml
```diff
@@ -298,7 +298,7 @@ ingress:
         host: '{{ .Release.Name }}.keycloak.example.com'
         # Paths for the host
         paths:
-          - path: /auth/admin/
+          - path: '{{ trimSuffix "/" .Values.http.relativePath}}/admin'
```
PSanetra, line 301: You might also need to adjust the following line https://github.com/codecentric/helm-charts/blob/d14250633937796653405b12c5699bf6362b7d18/charts/keycloakx/templates/ingress.yaml#L100 to look like `{{ tpl .path $ | quote }}`. Otherwise this template syntax will just be rendered as a string.

charts/keycloakx/values.yaml
```diff
@@ -403,7 +403,7 @@ serviceMonitor:
   # Timeout for scraping
   scrapeTimeout: 10s
   # The path at which metrics are served
-  path: /auth/metrics
+  path: '{{ trimSuffix "/" .Value.http.relativePath}}/metrics'
```
PSanetra, line 406: Please also update https://github.com/codecentric/helm-charts/blob/d14250633937796653405b12c5699bf6362b7d18/charts/keycloakx/templates/servicemonitor.yaml#L34 to render this string as a template.

PSanetra:
@dro-sh It would be nice if you could update the `values.schema.json` to include the `http.relativePath` option.

charts/keycloakx/values.yaml
dro-sh: Fixed

charts/keycloakx/values.yaml
dro-sh: Fixed

charts/keycloakx/values.yaml
dro-sh: Agree, in comment it looks strange.

charts/keycloakx/values.yaml
dro-sh: Fix

thomasdarimont:
Thanks! Merged

PR #577, opened 2022-04-19
fix ingress apiVersion with KubeVersion

e-genov-ks:
Fix ingress api version for Kubernetes version below  1.19+
According to the documentation of Kubernetes 
```
networking.k8s.io/v1beta1 = 1.14 to 1.18
networking.k8s.io/v1 = 1.19+
```

Similar with https://github.com/codecentric/helm-charts/pull/501

### DIFF
--- a/charts/keycloakx/templates/ingress.yaml
+++ b/charts/keycloakx/templates/ingress.yaml
@@ -1,6 +1,15 @@
 {{- $ingress := .Values.ingress -}}
 {{- if $ingress.enabled -}}
-{{- $apiVersion := "networking.k8s.io/v1" -}}
+{{- $apiV1 := false -}}
+{{- $apiVersion := "extensions/v1beta1" -}}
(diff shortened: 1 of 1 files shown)


PSanetra:
@e-genov-ks I am not sure if we should support Kubernetes < 1.19 on the new KeycloakX Chart. Kubernetes 1.18 is end of life since May 2021: https://kubernetes.io/releases/patch-releases/#support-period

e-genov-ks:
> @e-genov-ks I am not sure if we should support Kubernetes < 1.19 on the new KeycloakX Chart. Kubernetes 1.18 is end of life since May 2021: https://kubernetes.io/releases/patch-releases/#non-active-branch-history

Yes i know that. But this is still the only problem, everything else works fine on previous versions. Also there is a same fix  in keycloak helm chart. We can fix that behaviour or need to add info into README about minimal requirements  :(

PSanetra:
@thomasdarimont @mircohaug What do you think? Should we add a note about supported Kubernetes versions in the README files? Maybe to support only actively maintained versions?

mircohaug:
As @PSanetra already wrote I also think we should not support EOL Kubernetes versions.

PR #578, opened 2022-04-19
fix ClusterRole and ClusterRoleBinding names. Need if multiple namesp…

e-genov-ks:
If you have multiple namespaces, an error occurs:

`Error: rendered manifests contain a resource that already exists. Unable to continue with install: ClusterRole \"jgroups-kubeping-pod-reader\" in namespace \"\" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key \"meta.helm.sh/release-namespace\" must equal \"ns-temp-1\": current value is \"ns-temp-2\"`

### DIFF
--- a/charts/keycloakx/templates/serviceaccount.yaml
+++ b/charts/keycloakx/templates/serviceaccount.yaml
@@ -23,7 +23,7 @@ imagePullSecrets:
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: jgroups-kubeping-pod-reader
+  name: jgroups-kubeping-pod-reader-{{ .Release.Namespace }}
 rules:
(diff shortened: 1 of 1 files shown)


thomasdarimont:
Thanks for your contribution!

thomasdarimont (approved):
LGTM

PSanetra:
@thomasdarimont I think we should better replace the `ClusterRole` and `ClusterRoleBinding` with `Role` and `RoleBinding` as those are the namespaced resources, right?

thomasdarimont:
@PSanetra okay this makes sense. Will this cause problems for migration?

PSanetra:
@thomasdarimont I don't think so. The Role is just intended for discovery puposes between the Keycloak nodes, right?

PR #582, opened 2022-04-21
Tpl for relative path

ryshoooo:
Generalized the relative path for the tpl function. This is mainly in order to allow the relative path value to be referenced by other values or just for the user to do whatever they like for the relative path.

The relative path is extremely useful as it allows the Keycloak to be served on a prefix super easily. In my case, the prefix is completely dynamic based on external configuration, so being able to dynamically set the prefix inside this chart is very much wanted. This PR tries to accomplish that.


mircohaug:
Hi @ryshoooo can you explain your usecase in more detail. As I understand it if you set a custom value like `--set customvalue.prefix=foo` and then reference this custom value only within the prefix, then you can also set this prefix value directly. Correct me if I'm wrong.

ryshoooo:
> Hi @ryshoooo can you explain your usecase in more detail. As I understand it if you set a custom value like `--set customvalue.prefix=foo` and then reference this custom value only within the prefix, then you can also set this prefix value directly. Correct me if I'm wrong.

Hi @mircohaug sure I can explain. The prefix is in a configmap inside the cluster, so I'd like to do `--set customvalue.prefix={{ (lookup "v1" "ConfigMap" .Release.Namespace "config-values").data.KEYCLOAK_PREFIX  }}`, this way I can set the prefix dynamically without touching the keycloak chart values file. This variable KEYCLOAK_PREFIX is used in my other applications inside the cluster which communicate with keycloak, that's why I have it present as a configmap.

I could bypass this completely by overriding the "command", which is what I was doing before with the legacy keycloak helm chart. But it's ugly, I'd have to override health and liveness probes as well, etc. Considering that most of the values in this chart uses tpl, I thought this would be a more consistent and easier solution :)

mircohaug:
If you could squash your commits and rebase on master i will merge this.

PR #584, opened 2022-04-26
add namespace attribute in keycloakx objects

bovy89:
Signed-off-by: bovy89 <bovy89@gmail.com>


PSanetra:
Hi @bovy89, can you tell us more about the use case for this change? I think at least the changes related to setting the namespace fields to `.Release.Namespace` seem to be unnecessary.

bovy89:
Some tools are using helm template (`helm template -n <namespace>`) to generate a static manifest. That change is also helpful to include namespace info in order to exclude an entire namespace from automated testing (we are using trivy rego rules) without using anything else (i.e. kustomize).

Some honorable mentions of charts that are doing that:
- https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack/templates/prometheus
- https://github.com/kubernetes/ingress-nginx/tree/main/charts/ingress-nginx

charts/keycloakx/templates/prometheusrule.yaml
```diff
@@ -4,6 +4,9 @@ apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
   name: {{ include "keycloak.fullname" $ }}
+  {{- with .namespace }}
+  namespace: {{ . }}
```
mircohaug, line 8: This needs to default explicitly to `{{ .Release.Namespace }}` as well

mircohaug (changes requested):
Thanks for the contribution, please also add `{{ .Release.Namespace }}` as default when for the resources where namespace overwrites are possible. Namely the `PrometheusRule` you added the overwrite for but also the `ServiceMonitor` resource where this overwrite already exists

bovy89:
Any update?

grieshaber:
hey @bovy89 thanks for this PR. We had some changes in the maintainer team recently and lost a little bit of track.
I'll check your chages real quick and merge the PR if nothing comes up.

grieshaber (approved):

PR #587, opened 2022-05-05
feat(values): Add extraDeploy attribute to values

germainlefebvre4:
Allow to deploy out-of-the-box kubernetes manifests.


mircohaug:
Hi @germainlefebvre4, 

thanks for the contribution. I think the use-case you are describing here should be better implemented by using the [dependencies](https://helm.sh/docs/helm/helm_dependency/) feature of helm. You could place your extra microservice / configmap / etc in a dedicated chart and add a dependency on this chart. This also brings the upside, that you could manipulate the values passed to the keycloakx chart within the outer chart directly and don't rely on convention (e.g. for the configmap name). Also would this enable you better test and distribute this composite chart. 

If I misunderstood your usecase you disagree you are welcome bring forth additional arguments and reopen the PR.

Cheers Mirco

PR #592, opened 2022-05-11
[keycloakx] Major version upgrade to Keycloak 18.0.0 (#591)

### LINKED ISSUE #591: [keycloakx] Update to Keycloak 18.0.0

thomasdarimont:
Closes #591

Signed-off-by: Thomas Darimont <thomas.darimont@googlemail.com>

### DIFF
--- a/charts/keycloakx/examples/postgresql-kubeping/readme.md
+++ b/charts/keycloakx/examples/postgresql-kubeping/readme.md
@@ -2,7 +2,7 @@
 
 This example shows how to use KUBE_PING JGroup cluster discovery with Keycloak.X.
 
-Since Keycloak.X currently (17.0.1) does not support jgroups KUBE_PING out of the box,
+Since Keycloak.X (17.0.x, 18.0.x) does not support jgroups KUBE_PING out of the box,
 we need to download the library and copy it into a custom docker image.
 
 Note that we use some customizations in the `keycloak-server-values.yaml` file:
--- a/charts/keycloakx/templates/statefulset.yaml
+++ b/charts/keycloakx/templates/statefulset.yaml
@@ -126,6 +126,10 @@ spec:
             - name: KC_METRICS_ENABLED
               value: "true"
             {{- end }}
+            {{- if .Values.metrics.enabled }}
+            - name: KC_HEALTH_ENABLED
+              value: "true"
+            {{- end }}
             {{- with .Values.extraEnv }}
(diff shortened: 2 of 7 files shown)


charts/keycloakx/examples/postgresql-kubeping/readme.md
```diff
@@ -2,7 +2,7 @@
 
 This example shows how to use KUBE_PING JGroup cluster discovery with Keycloak.X.
 
-Since Keycloak.X currently (17.0.1) does not support jgroups KUBE_PING out of the box,
+Since Keycloak.X currently (18.0.0) does not support jgroups KUBE_PING out of the box,
```
PSanetra, line 5: Should we leave the old version number to avoid confusion?

charts/keycloakx/templates/statefulset.yaml
```diff
@@ -126,6 +126,10 @@ spec:
             - name: KC_METRICS_ENABLED
               value: "true"
             {{- end }}
+            {{- if .Values.metrics.enabled }}
+            - name: KC_HEALTH_ENABLED
+              value: "true"
```
PSanetra, line 131: Is there a reason to disable this check? I am thinking about using these health check endpoints for the readiness, startup and liveness probes. Or is the current approach for those probes still more appropriate?

thomasdarimont: We currently use the `/auth` / `/auth/realms/master` as readiness and liveness probes.

This health endpoint just exposes the `/auth/health` endpoint that can be customized.
The keycloak health endpoint is disabled by default, see: https://www.keycloak.org/server/all-config#_health

PSanetra: I was just wondering about the purpose of the health endpoint and thought it would be most useful for readiness and liveness probe use cases. (Especially as it seems to support `/health/ready` and `/health/live` explicitly, but I don't know what those endpoints really do.)

ThorstenKunz:
Any chance to get this merged to get the chart pushed to kc 18?

KlavsKlavsen:
I would love to get this merged too :)

charts/keycloakx/templates/statefulset.yaml
grieshaber: I'd say, we stick to the "old" readiness and livenessProbes for now  and discuss switching to the /health endpoint in another thread.

PR #600, opened 2022-05-17
doc (keycloakx): Add example with ingress (#599)

### LINKED ISSUE #599: [keycloakx] Add example with ingress

thomasdarimont:
Closes #599

Signed-off-by: Thomas Darimont <thomas.darimont@googlemail.com>

### DIFF
--- a/charts/keycloakx/examples/postgresql-ingress/keycloak-server-values.yaml
+++ b/charts/keycloakx/examples/postgresql-ingress/keycloak-server-values.yaml
@@ -1,0 +1,78 @@
+# This is an example configuration, for production grade configuration see the Keycloak documentation.
+# See https://www.keycloak.org/server/configuration
+# See https://www.keycloak.org/server/all-config
+command:
+  - "/opt/keycloak/bin/kc.sh"
+  - "--verbose"
+  - "start"
+  - "--auto-build"
+  - "--http-enabled=true"
+  - "--http-port=8080"
+  - "--hostname-strict=false"
+  - "--hostname-strict-https=false"
+  - "--spi-events-listener-jboss-logging-success-level=info"
+  - "--spi-events-listener-jboss-logging-error-level=warn"
+  - "--proxy=edge"
+
+# Configure ingress
+ingress:
+  enabled: true
+  annotations:
+    kubernetes.io/ingress.class: nginx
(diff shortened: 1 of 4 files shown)


charts/keycloakx/examples/postgresql-ingress/keycloak-server-values.yaml
```diff
@@ -0,0 +1,78 @@
...
+  - "--hostname-strict-https=false"
+  - "--spi-events-listener-jboss-logging-success-level=info"
+  - "--spi-events-listener-jboss-logging-error-level=warn"
+  - "--proxy=edge"
+
+# Configure ingress
+ingress:
+  enabled: true
+  annotations:
+    kubernetes.io/ingress.class: nginx
```
grieshaber, line 21: Shouldn't we use the ingressClassName field instead of the "deprecated" annotation here?
see https://github.com/codecentric/helm-charts/blob/master/charts/keycloakx/templates/ingress.yaml#L21

charts/keycloakx/examples/postgresql-ingress/keycloak-server-values.yaml
```diff
@@ -0,0 +1,78 @@
...
+      - id.acme.test
+      paths:
+        - path: /auth
+          pathType: Prefix
+  console:
+    enabled: true
+    tls:
+      - hosts: 
+        - id.acme.test
+        paths:
```
grieshaber, line 39: the `paths` blocks need to be below a `rules` block instead of `tls`
```suggestion
     rules:
       - hosts:
          - id.acme.test
         paths:
```

charts/keycloakx/examples/postgresql-ingress/keycloak-server-values.yaml
```diff
@@ -0,0 +1,78 @@
...
+    ingress.kubernetes.io/affinity: cookie
+  rules:
+    - host: id.acme.test
+      paths:
+        - path: /auth
+          pathType: Prefix
+  tls:
+    - hosts: 
+      - id.acme.test
+      paths:
```
grieshaber, line 31: see comment below

grieshaber (changes requested):

PR #601, opened 2022-05-18
use existing secret for  keycloakx database password

ismailyenigul:
This path allows us to use existing k8s secret to set database password instead of setting it in values.yml plain text.
This approach is good for GitOps not to keep secret in plaintext in git repo.
the secret can be created externally and save t encrypted(eg sealed-secrets) in git repo.
PS: I will push another commit for  README.md if you update it manually.

### DIFF
--- a/charts/keycloakx/templates/_helpers.tpl
+++ b/charts/keycloakx/templates/_helpers.tpl
@@ -71,9 +71,7 @@ Create the service DNS name.
   valueFrom:
     secretKeyRef:
     {{- if .Values.database.existingSecret }}
-      name: {{ .Values.database.existingSecret }}
-    {{- else }}
-      name: {{ template "keycloak.fullname" . }}-database
+      name: {{ .Values.database.existingSecret | default (printf "%s-database" (include "keycloak.fullname" . ))}}
     {{- end }}
       key: password
   {{- end }}
--- a/charts/keycloakx/templates/_helpers.tpl
+++ b/charts/keycloakx/templates/_helpers.tpl
@@ -64,3 +64,17 @@ Create the service DNS name.
 {{- define "keycloak.serviceDnsName" -}}
 {{ include "keycloak.fullname" . }}-headless.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
 {{- end }}
+
+{{- define "keycloak.databasePasswordEnv" -}}
+{{- if or .Values.database.password .Values.database.existingSecret -}}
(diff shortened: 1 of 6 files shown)


charts/keycloakx/templates/_helpers.tpl
```diff
@@ -64,3 +64,17 @@ Create the service DNS name.
...
 {{ include "keycloak.fullname" . }}-headless.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
 {{- end }}
+
+{{- define "keycloak.databasePasswordEnv" -}}
+{{- if or .Values.database.password .Values.database.existingSecret -}}
+- name: KC_DB_PASSWORD
+  valueFrom:
+    secretKeyRef:
+    {{- if .Values.database.existingSecret }}
+      name: {{ .Values.database.existingSecret }}
```
grieshaber, line 74:
```suggestion
      name: {{ .Values.database.existingSecret | default (printf "%s-database" (include "keycloak.fullname" . ))}}
```
and drop the rest of the if clause below.

charts/keycloakx/values.yaml
```diff
@@ -364,6 +364,8 @@ dbchecker:
       memory: "32Mi"
 
 database:
+  # don't create secret for db password. Instead use existing k8s secret
+  # existingSecret: "my-existent-dbpass-secret"
```
grieshaber, line 368:
```suggestion
  # existingSecret: "my-existent-dbpass-secret"
  existingSecret: ""
```
to avoid nil pointer

grieshaber (changes requested):
Thanks for your PR - i made some suggestions with small "improvements".

Regarding all the release relevant stuff - as mentioned at the changes, please revert them, as i will take care of an cumulated release later on.

grieshaber:
@ismailyenigul could also please make sure, that your commits contain a valid signature? This is needed to check the DCO action

ismailyenigul:
sorry for new releases. i had to test it in my env first. now going to apply changes.

grieshaber:
no worries, in general it is fine to do so. But, as we just need to catch up our maintenance of this repo, things are a bit "more manual" than needed.

Thanks for the fast updates of this PR! I see you resolved the discussion and added a commit message with "added suggested changes" - but, if i'm not mistaken, the commit did not contain the suggested changes.

ismailyenigul:
just pushed new commit.
Please check

grieshaber (approved):

PR #623, opened 2022-07-20
fix: correct fallback when no existingSecret

wilmardo:
Followup of: #601 

The previous setup resulted in the `name` missing when no existingSecret is provided:

```yaml
 name: KC_DB_PASSWORD
  valueFrom:
    secretKeyRef:
      key: password
```

Since `{{- if .Values.database.existingSecret }}` is then false and the whole thing is skipped.


grieshaber:
Thanks @wilmardo for the catch - it slipped through in #601, as some improvements where added in the process. I'll release the fix as soon as the checks are completed.

PR #626, opened 2022-07-22
feature: Allow setting mailhog replicas

Jancis:
- Sets mailhog `spec.replicas` value, allows override via helm values
- Upstream update to `5.0.7`
- Uses `podReplicas` instead of `replicaCount` to match existing `podXYZ` variables

PR based on https://github.com/codecentric/helm-charts/pull/206 , credits to @bartlettt

### DIFF
--- a/charts/mailhog/values.yaml
+++ b/charts/mailhog/values.yaml
@@ -94,6 +94,8 @@ outgoingSMTP:
     #   host: "mail2.example.com"
     #   port: "587"   # NOTE: go requires this port number to be a string... otherwise the container won't start
 
+podReplicas: 1
+
 podAnnotations: {}
 
 podLabels: {}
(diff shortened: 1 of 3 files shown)


charts/mailhog/values.yaml
```diff
@@ -94,6 +94,8 @@ outgoingSMTP:
     #   host: "mail2.example.com"
     #   port: "587"   # NOTE: go requires this port number to be a string... otherwise the container won't start
 
+replicaCount: 1
```
bartlettt, line 97: should be `podReplicas`

Jancis: Wow, that was swift response!
had 3 lines, made error in one of them :facepalm:

Jancis:
I signed commit but DCO check does not re-trigger for some reason.

grieshaber:
@Jancis thanks for the PR. As far as i can see, your commit is verified, but not signed-off as DCO expects it to.
Then it would have a "signed-off by: author author@mail.com" in the commit message, which your commit has not.
You can achieve this by adding -s to the `git commit` command. 
Could you please update & force push?

grieshaber:
PR is merged and will be release tomorrow afternoon as a new feature version.

PR #627, opened 2022-07-29
[mailhog] Add extraPorts in Service and extraContainers in Deployment

jullianow:
When deploying the mailhog graph using ingress GCE the health check of Google's Load Balancer fails to succeed if the service is configured with basic auth.

This is because the health check is not able to set authentication parameters so that a request gets a return code of 200.
Also, even though there is an smtp port, the load balancer health check only has the ability to communicate via HTTP, HTTPS and HTTP2 protocol.
Reference: https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-features#direct_health

Therefore, the purpose of this change is to give the ability to add an extra container to be used only as a route for the Load Balancer life check to work, at the same time that mailhog basic authentication is enabled.

### DIFF
--- a/charts/mailhog/README.md
+++ b/charts/mailhog/README.md
@@ -63,7 +63,7 @@ Parameter | Description | Default
 `service.annotations` | Annotations for the service | `{}`
 `service.clusterIP` | Internal cluster service IP | `""`
 `service.externalIPs` | Service external IP addresses | `[]`
-`service.extraPorts` | Additional ports to be added service | `[]`
+`service.extraPorts` | Additional ports to the service | `[]`
 `service.loadBalancerIP` | IP address to assign to load balancer (if supported) | `""`
 `service.loadBalancerSourceRanges` | List of IP CIDRs allowed access to load balancer (if supported) | `[]`
 `service.type` | Type of service to create | `ClusterIP`
--- a/charts/mailhog/README.md
+++ b/charts/mailhog/README.md
@@ -40,6 +40,7 @@ The following table lists the configurable parameters of the Mailhog chart and t
 
 Parameter | Description | Default
 --- | --- | ---
+`extraContainers` | Additional containers to be added to the application pod | `[]`
 `image.repository` | Docker image repository | `mailhog/mailhog`
 `image.tag` | Docker image tag whose default is the chart version | `""`
 `image.pullPolicy` | Docker image pull policy | `IfNotPresent`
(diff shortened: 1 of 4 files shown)


charts/mailhog/README.md
```diff
@@ -62,6 +63,7 @@ Parameter | Description | Default
 `service.annotations` | Annotations for the service | `{}`
 `service.clusterIP` | Internal cluster service IP | `""`
 `service.externalIPs` | Service external IP addresses | `[]`
+`service.extraPorts` | Additional ports to be added service | `[]`
```
grieshaber, line 66: Reads like there is missing sth - added `to` the service e.g.?

charts/mailhog/templates/deployment.yaml
```diff
@@ -86,6 +86,9 @@ spec:
           {{- end }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+      {{- with .Values.extraContainers }}
+        {{- toYaml . | nindent 6 }}
```
grieshaber, line 90:
```suggestion
        {{- toYaml . | nindent 8 }}
```
if i'm not mistaken, we need an nindent of `8` instead of `6`. With your proposal, i cannot template the mailhog chart anymore.

grieshaber (changes requested):

grieshaber (approved):

jullianow:
@grieshaber Very good.

jullianow:
@grieshaber I'm sending the message in this thread because it has to do with it.

Do you think a PR adding a route (E.g. /health) to the Mailhig service makes sense?
This route would not have configuration for basic authentication, so it could be used in the Ingress GCE health check.

That's because, even adding a specific container to be used in the health check (as I described in this PR) is not enough.
Google tries to ensure that [all ports exposed by Ingress](https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-features#unique_backendconfig_per_service_port) are returning 200, and therein lies the problem.
When Mailhog has basic authentication, the check fails.
image

jullianow:
Yes, it is possible to do what you are talking about.
But the problem is that I can define which backendconfig should be used for a port.
But the point is that the BackendConfig (either mine or the default that exists naturally) it validates all ports that are specified in Ingress (which is the image I sent).

That is, this suggestion of yours was exactly what I did and that I verified that it doesn't work (after seeing the information block of the official documentation).

>But, to understand you 2nd proposal further:
You want to add a valid /health endpoint to mailhog itself and use this for the readnessProbe (which google will then use aswell)?

Perfect. And this route will not suffer from the basic authentication configuration.
So I can have a BackendConfig where the healthCheck will have a requestPath for this route and everything will work.

PR #629, opened 2022-08-01
[mailhog] Add ability to precisely control liveness and readiness probes

jullianow:
In some use cases we need to have more control over the liveness probes on standby so that the container is available at the right time.
Also, in some cases, probes that vary the process in a short time, can produce many logs that we do not necessarily want.
Therefore, this change in the ability to set all existing properties to more precisely control with the liveliness and readiness probes should behave.

> **Note**
>
>All default values ​​are based on the documentation and what was already in use by the chart.

### DIFF
--- a/charts/mailhog/README.md
+++ b/charts/mailhog/README.md
@@ -56,6 +56,16 @@ Parameter | Description | Default
 `resources` | Pod resource requests and limits | `{}`
 `tolerations` | Node taints to tolerate | `[]`
 `priorityClassName` | Name of the existing priority class to be used by Mailhog pod, priority class needs to be created beforehand | `""`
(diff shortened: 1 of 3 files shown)


grieshaber:
Thanks for the PR @jullianow.

I'm not sure, if i like the very explicit setting of the liveness and readnessprobe and / or having the values explicitly predefined.

In order to achieve flexibility, i'd at least use sth like this in the deployment.yaml:
`
        readinessProbe: {{ toYaml .Values.readinessProbe | nindent X }}
        livenessProbe: {{ toYaml .Values.livenessProbe | nindent X }}
`
So, this way, if new or other properties are added, we can make use of them directly.

grieshaber (approved):

PR #632, opened 2022-08-02
[mailhog] Add ability to definit targetPort of name based service or not

jullianow:
Named target ports are not supported by GCE health checks, so when deploying in GKE and exposing it via GCE ingress, the health checks fail and the load balancer returns a 502.


jullianow:
Ping

charts/mailhog/values.yaml
```diff
@@ -108,13 +119,13 @@ extraEnv: []
 # ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes
 livenessProbe:
   tcpSocket:
-    port: 1025
+    port: *containerSmtpPort
```
grieshaber, line 122: I see, what you try to achieve here - but imho this is not considered best practice to use the yaml anchors with helm.

i'd say, tackle the behaviour you want to achieve in the templates directly - so skip the definition in the _defaults_ here and access containerPort in the template or livenessProbe Port when defined.

grieshaber (changes requested):

jullianow:
Ping

jullianow:
Ping

jullianow:
@grieshaber Can we walk with this PR?

grieshaber:
Hey @jullianow, 
the release will be created automatically tomorrow afternoon. (Most likely not following semvar best-practices, as the release mechanism is not correctly used and enforced unfortunately).

jullianow:
Sorry, but I don't think I understand.  The graph version will not be generated following the semver-based version evocation pattern, but a version will be generated following another pattern?

jullianow:
@grieshaber I have a curiosity.  Do you know or maintain the Mailhog application repository?  Do you have any news why the PRs there are “half stopped”?

PR #633, opened 2022-08-03
[keycloakx] Bump Keycloak to v19.0.2

Footur:
I've updated the Keycloak container image to version 19.0.2.


igor-nikiforov:
It make sense to change path for readiness and liveness probes according upgrade guide as well https://www.keycloak.org/docs/latest/upgrading/index.html#potentially-breaking-changes-to-the-health-endpoints.

Footur:
@igor-nikiforov Thank you for pointing this out. I've changed the default values for the self probes. The chart version will be changed automatically, according to my understanding.

grieshaber:
@Footur regarding the change of readness and liveness probe, i think we should tackle https://github.com/codecentric/helm-charts/issues/636 in the same release. I'll try to add a PR for this in the next days.

grieshaber:
Currently "blocked" by #638

grieshaber:
Reading through the release-notes, i read that a flag we have in the /examples `--auto-build` is now deprecated, so i'd like to mention the deprecation there and ideally make not use of it anymore.
Suggestion for file `charts/keycloakx/examples/*/keycloak-server-values.yaml`:
```
command:
  - "/opt/keycloak/bin/kc.sh"
  - "--verbose"
  - "start"
  # - "--auto-build" deprecated and now "default"
  # - "--optimized" disable new default with this flag
```

Footur:
I've removed the `--auto-build` flag. According to my understanding, to use the `--optimize` option, `kc.sh build` must be run first.
[Optimize the Keycloak startup](https://www.keycloak.org/server/configuration#_optimize_the_keycloak_startup).

Does the lint test fail because the UI of Keycloak has changed?

grieshaber:
Agreed, i misread that section.

Regarding lint test:

It seems, that it now tries to access /auth/health/ endpoints. At least if i understood it correctly, the /auth/ path is wrong here (for the healthchecks).
Therefore the Pods wont become ready and b/c of that the test are failing.

Root-Issue is 
```
http:
  # For backwards compatibility reasons we set this to the value used by previous Keycloak versions.
  relativePath: "/auth"
```
which is used in the probes.

Footur:
@grieshaber I've tested the realtive path with "/", but the lint test is still failing.

https://github.com/codecentric/helm-charts/runs/7688193943?check_suite_focus=true

grieshaber:
I see the following issue:
```
"http://10.244.1.2:8080/health": dial tcp 10.244.1.2:8080: connect: connection refused
```

Footur:
Unfortunately, I can reproduce the error in my local setup. I also don't have the time to delve deeper into the CI because I'm on holiday for a two weeks starting tomorrow.

grieshaber:
I really appreciate your effort @Footur. I'll check on my local setup as well and if i find sth will update here.

grieshaber:
So, the issue it not connected to startup probe etc. The pod spin up without any issue.
But the executed tests are failing - i will now try to understand why and if there is an obvious part that broke with the update.

Footur:
@grieshaber I've fixed the deployment, but the Selenium test is failing.

grieshaber:
@Footur yes, i saw that as well.. from selenium pov the login / redirect to login UI is not working properly. I started debugging some time ago and had issues aswell - sometimes it just went on and on loading forever and sometimes it worked immediately. But i did not make it work in the UI tests..

aleshchynskyi:
Is there any chance it will be released in the nearest future (a week?)?

Footur:
> Is there any chance it will be released in the nearest future (a week?)?

Hello Anton,

I don't have enough time right now to figure out the root of the problem. In our development environment the chart is working. Feel free to commit by your self. 

Cheers,
Footur

cocor-au-lait:
Hello. I'm looking forward to this update!
About test failure, I also tried to find out

The first thing I tried was to disable the new admin console that became the default in 19.x.

`charts/keycloakx/ci/h2-values.yaml/h2-values.yaml`
```yaml
extraEnv: |
  - name: KC_FEATURES_DISABLED
    value: admin2
```

The test finished successfully.

I wonder why the tests failed on the new admin console.
(like: PhatomJS not compatible with newer JS frameworks?)

---

I think tests that disable default features are unreliable.
So based on @Footur's method of using ChromeDriver, I tried to use different Docker image.

```yaml
test:
  image:
    repository: docker.io/joyzoursky/python-chromedriver
    tag: 3.9-selenium
```
https://github.com/joyzoursky/docker-python-chromedriver

The test finished successfully in new admin console!

See below for final changes.

https://github.com/cocor-au-lait/keycloak-helm-charts/pull/1

If you don't mind, I would appreciate it if you could reflect this change in this PR.

charts/keycloakx/ci/h2-values.yaml
```diff
@@ -1,7 +1,7 @@
 command:
   - "/opt/keycloak/bin/kc.sh"
   - "--verbose"
-  - "start-dev"
+  - "start"
```
cocor-au-lait, line 4: I think the test is failing because of booting in production mode.
It is necessary to run in production mode for testing?

cocor-au-lait: I just created test that launch in production mode and passes.

https://github.com/cocor-au-lait/keycloak-helm-charts/pull/3/files#diff-a2787f787ce9c889ac6c604daa1c10fce6d02e7478333378f82bf541863a0b70R22-R25

To pass the test, I've disabled strict hostname and https.

Footur:
Hey @cocor-au-lait, Thank you for the contribution.

> I wonder why the tests failed on the new admin console.
> (like: PhatomJS not compatible with newer JS frameworks?)

I've change PhantomJS to Chrome, because PhantomJS is [deprecated](https://github.com/ariya/phantomjs/issues/15344).



>  just created test that launch in production mode and passes.
> 
> https://github.com/cocor-au-lait/keycloak-helm-charts/pull/3/files#diff-a2787f787ce9c889ac6c604daa1c10fce6d02e7478333378f82bf541863a0b70R4-R5
>
>To pass the test, It's need to set strict hostname and https.

I don't see a change in the test behavior.

charts/keycloakx/ci/h2-values.yaml
```diff
@@ -20,6 +22,10 @@ extraEnv: |
       -XX:MaxRAMPercentage=50.0
       -Djava.awt.headless=true
       -Djgroups.dns.query={{ include "keycloak.fullname" . }}-headless
+  - name: KC_HOSTNAME
+    value: keycloak-ci
```
cocor-au-lait, line 26: > I don't see a change in the test behavior.

```suggestion
```

`KC_HOSTNAME` is the reason why the test fails.

charts/keycloakx/ci/h2-values.yaml
```diff
@@ -20,6 +22,10 @@ extraEnv: |
...
+  - name: KC_FEATURES_DISABLED
+    value: admin2
```
cocor-au-lait, line 28:
```suggestion
```

Testing the new admin console with the Chrome driver now passes, so I think it's better to test with default behavior.

cocor-au-lait: https://github.com/codecentric/helm-charts/blob/cab3997511239923a030333e006db569239d2ca3/charts/keycloakx/README.md?plain=1#L11-L25

I think it's more reliable to define the test manifest according to the guidance in the `README`.

charts/keycloakx/README.md
```diff
@@ -205,7 +204,7 @@ The following table lists the configurable parameters of the Keycloak-X chart an
...
+| `test.image.repository`                      | The image for the test Pod                                                                                                                                                                                                                                                        | `docker.io/joyzoursky/python-chromedriver`                                                                              |
 | `test.image.tag`                             | The tag for the test Pod image                                                                                                                                                                                                                                                    | `v1`                                                                                                                                  |
```
cocor-au-lait, line 208:
```suggestion
| `test.image.tag`                             | The tag for the test Pod image                                                                                                                                                                                                                                                    | `3.9-selenium`                                                                                                                                  |
```

charts/keycloakx/ci/h2-values.yaml
```diff
@@ -32,5 +38,31 @@ secrets:
       user: admin
       password: secret
 
+http:
+  relativePath: "/"
```
cocor-au-lait, line 42: I've confirmed that even default path `/auth` also pass the test.

https://github.com/codecentric/helm-charts/blob/cab3997511239923a030333e006db569239d2ca3/charts/keycloakx/values.yaml#L394-L396

So I think it's better not to set it individually.

Footur:
@cocor-au-lait Thank you for the hint. It's fixed now. 💖

@grieshaber Can you review the PR and provide some feedback please? I've changed the selenium container image, because PhantomJS is [deprecated](https://github.com/ariya/phantomjs/issues/15344).

grieshaber:
@Footur @cocor-au-lait thank you both for resolving the issue(s)! I'll do the review until the next scheduled release, so that we have the new version release friday afternoon!

charts/keycloakx/templates/statefulset.yaml
```diff
@@ -156,10 +156,12 @@ spec:
           livenessProbe:
             {{- tpl . $ | nindent 12 }}
           {{- end }}
+          {{- if .Values.metrics.enabled }}
```
tomasz-baran, line 159: Does this line forces enabling metrics in order to have any readinessProbe?
If yes, IMO it would reduce flexibility of defining a custom command based probe.

For me info in documentation is sufficient for those who want to use Keycloak health checks :)

Furthermore, the endpoint is still active even when metrics are disabled:

1) metrics enabled:
```
bash-4.4$ curl -k https://127.0.0.1:8443/health/ready
{
    "status": "UP",
    "checks": [
        {
            "name": "Keycloak database connections health check",
            "status": "UP"
        }
    ]
}
```
2) metrics disabled:
curl -k https://127.0.0.1:8443/health/ready
{
    "status": "UP",
    "checks": [
    ]
}

It's not clear for me whether I really need the `Database` check enabled for readiness checking:
https://www.keycloak.org/server/health

And since now we have the `Database` check maybe the `dbchecker` functionality is not needed anymore and could be deprecated.

charts/keycloakx/README.md
```diff
@@ -174,7 +173,7 @@ The following table lists the configurable parameters of the Keycloak-X chart an
...
-| `health.enabled`                            | If `true` then the health endpoint is exposed                                                                                                                                                                                                                                    | `true`                                                                                                                                |
+| `health.enabled`                            | If `true` then the health endpoint is exposed. If the `readinessProbe` is is needed `metrics.enable` must be `true`.                                                                                                                                                                                                                                    | `true`                                                                                                                                |
```
tomasz-baran, line 176: > If the `readinessProbe` is is needed `metrics.enable` must be `true`
@Footur are you sure this is true? Asking since Keycloak Operator creates readinessProbe with metrics disabled. For some reason, by default, metrics are disabled. Currently I have metrics disabled but defined readinessProbe. After this change I will be forced to enable the metrics to not lose the probe.

Footur:
@grieshaber I hope you have some time to review this pull request soon. :)

charts/keycloakx/templates/statefulset.yaml
grieshaber: @Footur you marked this discussion as resolved. @tomasz-baran pointed out a valid issue here i think - we reduce the flexibility here; maybe we should just drop the line here, but keep the information in the readme. This way one can configure the probe as needed (for whatever workaround one might want to implement).

charts/keycloakx/values.yaml
```diff
@@ -500,9 +500,9 @@ test:
   enabled: false
   image:
     # The image for the test Pod
-    repository: docker.io/unguiculus/docker-python3-phantomjs-selenium
+    repository: docker.io/joyzoursky/python-chromedriver
```
grieshaber, line 503: The project / image does not seem to be very "maintained". But, as we replace an ever less maintained image, this is fine for now..

charts/keycloakx/templates/statefulset.yaml
grieshaber: Additionally, i know, i requested the change initially. But with the issue @tomasz-baran pointed out, my "old" request might be not valid..

charts/keycloakx/values.yaml
cocor-au-lait: I agree, but it seems that there is no image that continues to be maintained, so it would be better to createtest image in the repository in the future.

grieshaber:
@Footur seems like this PR needs a rebase before merging aswell..

Footur:
@grieshaber What makes you think that? This branch is up-to-date.

charts/keycloakx/templates/statefulset.yaml
tomasz-baran: @grieshaber Thanks for the comments. I've checked Keycloak code and you are right. The `ready` check is defined in `KeycloakReadyHealthCheck` class which extends `DataSourceHealthCheck`. There is also PR with design doc with 
[a comment](https://github.com/keycloak/keycloak-community/pull/55/files#r528560159) on this topic.

I was misleading by the Keycloak Operator code. However, it looks like Keycloak project wants to not force enabling metrics in order to use the `/health/ready` endpoint - based on their test code:
```
    @Test
    @Launch({ "start-dev", "--health-enabled=true" })
    void testHealthEndpoint() {
        when().get("/health").then()
                .statusCode(200);
        when().get("/health/live").then()
                .statusCode(200);
        when().get("/health/ready").then()
                .statusCode(200);
        // Metrics should not be enabled
        when().get("/metrics").then()
                .statusCode(404);
    }
```

grieshaber:
> @grieshaber What makes you think that? This branch is up-to-date.

nvm, Github was per default set to rebase - which is not possible due to conflicts. Merge is possible..

charts/keycloakx/templates/statefulset.yaml
grieshaber: Ok - i'd say, let's stick to the current implementation, as this implies the most valid readiness check for the software the chart chips. If another use-case arises, we can adapt again to that use-case.

grieshaber (approved):

Issue #636, opened 2022-08-04
[keycloakx] Bug: health check depends on metrics.enabled flag


grieshaber:
Instead of reliying on the newly introduced `health.enabled` (https://github.com/codecentric/helm-charts/pull/592/files) flag, the HEALTH endpoint is enabled via `metrics.enabled` flag. 

This might lead to confusion and should be fixed.

Footur:
@grieshaber I've published a pullrequest regarding this issue (#638).

grieshaber:
.. after reading the release notes of keycloak 19, i realized: relying on `metrics.enabled` was not totally wrong, but also not enough.
```
The health/ready endpoint used for readiness probes still checks for a working database connection. Make sure you have not only health-enabled=true but also metrics-enabled=true set in your configuration, to enable the database check, resulting in an effective readiness probe. It will return HTTP status-code 503 and a status of DOWN when the database connection is not in a healthy state.
```
So in order to have _fully_ functional readinessChecks, we need both, `health` and `metrics` endpoints enabled.

Footur:
@grieshaber Maybe we should write a note in the README?

grieshaber:
@Footur yes, that is definetly a good idea. I'm just thinking about a "technical" solution as well. As we have the if clause for health enabled in place, should we maybe add the check for health and metrics to the readnessProbe?

grieshaber:
Thanks to @Footur the latest idea (Readme + technical feature gate) will be resolved with https://github.com/codecentric/helm-charts/pull/633.

grieshaber:
Resolved.

PR #638, opened 2022-08-04
[keycloakx] Health check depens on health.enabled

Footur:
A potential fix for bug #636.

### DIFF
--- a/charts/keycloakx/templates/statefulset.yaml
+++ b/charts/keycloakx/templates/statefulset.yaml
@@ -127,7 +127,7 @@ spec:
             - name: KC_METRICS_ENABLED
               value: "true"
             {{- end }}
-            {{- if .Values.metrics.enabled }}
+            {{- if .Values.health.enabled }}
             - name: KC_HEALTH_ENABLED
               value: "true"
             {{- end }}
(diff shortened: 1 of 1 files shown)


charts/keycloakx/templates/statefulset.yaml
```diff
@@ -148,6 +148,7 @@ spec:
             {{- with .Values.extraPorts }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
+          {{- if .Values.health.enabled }}
```
grieshaber, line 151: I'd say we can skip this if clause - if someone decides to stick to the "old" livness-probes, he might not need to enable the health endpoint. But, with the proposal, he needs to enable the endpoint in order to use live-, readiness and startupProbe at all. What do you think?

grieshaber (changes requested):

charts/keycloakx/templates/statefulset.yaml
Footur: I don't see the use case here. Why should anyone want to use the old health checks?

Footur: The health endpoint is enabled by default. This PR is about using health checks at all. It doesn't matter in this case if the old health checks are configured or not.

grieshaber: In that matter correct - but there would be a dependency between `health.enabled` and the usage of livenessProbe etc. 
I agree, that the use case for not using health for healthchecks is very abstract and i cannot think of any valid reason.. so let's merge.

grieshaber (approved):

Issue #622, opened 2022-07-18
[keycloakx] PDB is not choosing between API versions


Tim-Schwalbe:
In the old keycloak helm chart there was a selection of the available API for PDB.

```
{{- if .Values.podDisruptionBudget -}}
{{- if .Capabilities.APIVersions.Has "policy/v1" }}
apiVersion: policy/v1
{{- else }}
apiVersion: policy/v1beta1
{{- end }}
```

In the new keycloakx template only this is used `apiVersion: policy/v1beta1`

https://github.com/codecentric/helm-charts/blob/master/charts/keycloakx/templates/poddisruptionbudget.yaml

```
{{- if .Values.podDisruptionBudget -}}
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  name: {{ include "keycloak.fullname" . }}
  labels:
    {{- include "keycloak.labels" . | nindent 4 }}
spec:
  selector:
    matchLabels:
      {{- include "keycloak.selectorLabels" . | nindent 6 }}
  {{- toYaml .Values.podDisruptionBudget | nindent 2 }}
{{- end -}}

```

grieshaber:
Resolved with #649

PR #649, opened 2022-09-19
use PDB policy/v1 if supported  (resolve issue #622)

bovy89:
Signed-off-by: bovy89 <bovy89@gmail.com>

Use policy/v1 if supported

### DIFF
--- a/charts/keycloakx/templates/poddisruptionbudget.yaml
+++ b/charts/keycloakx/templates/poddisruptionbudget.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.podDisruptionBudget -}}
-apiVersion: policy/v1beta1
+apiVersion: {{ ternary "policy/v1" "policy/v1beta1" (semverCompare ">=1.21.0-0" .Capabilities.KubeVersion.Version) }}
 kind: PodDisruptionBudget
 metadata:
   name: {{ include "keycloak.fullname" . }}


charts/keycloakx/templates/poddisruptionbudget.yaml
```diff
@@ -1,5 +1,6 @@
 {{- if .Values.podDisruptionBudget -}}
-apiVersion: policy/v1beta1
+{{- if or (and .Values.autoscaling.enabled (gt (.Values.autoscaling.minReplicas | int) 1)) (and (not .Values.autoscaling.enabled) (gt (.Values.replicas | int) 1)) }}
```
grieshaber, line 2: @bovy89 i'm not sure, if i like this restriction - at least without any part in the documentation describing this limitation.
Additionally, as the default is "no" PDB, i'd say it doesn't need to be double-checked here.

grieshaber (changes requested):

charts/keycloakx/templates/poddisruptionbudget.yaml
bovy89: That could be an additional safety measure because PDB with replica 1 is usually a wrong configuration. That is not strictly necessary...I can remove it if you want

grieshaber: I'd say, let's drop this step, as it introduces new layers of dependecies to values etc. The updated API-Version is indeed a good thing to fix! :)

grieshaber (approved):

PR #651, opened 2022-09-27
feat: only expose https ports when https port is defined in the service

pyromaniac3010:
The https service and https containerports are typically not needed as most implementations will use the ingress with http backend communication. The https ports should therefore be disabled by default and not included in the service and statefulset. Only if someone manually sets a https port number via values, it should be exposed.
This also fixes some default prometheusrules which alert if service endpoints are not available.

### DIFF
--- a/charts/keycloakx/templates/service-http.yaml
+++ b/charts/keycloakx/templates/service-http.yaml
@@ -42,6 +42,7 @@ spec:
       nodePort: {{ .Values.service.httpNodePort }}
       {{- end }}
       protocol: TCP
+      {{- if .Values.service.httpsPort }}
     - name: https
       port: {{ .Values.service.httpsPort }}
       targetPort: https
@@ -49,6 +50,7 @@ spec:
       nodePort: {{ .Values.service.httpsNodePort }}
       {{- end }}
       protocol: TCP
+      {{- end }}
     {{- with .Values.service.extraPorts }}
     {{- toYaml . | nindent 4 }}
     {{- end }}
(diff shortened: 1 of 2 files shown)


pyromaniac3010:
@grieshaber @mircohaug  Anything wrong with this PR?

pyromaniac3010:
@grieshaber the mailhog chart was accidentally committed. this was caused by a rebase on my side. 
you could leave the default port setting as it was to avoid a major release. but I think it should be changed to null with the next release as ssl termination is usually done at the ingress level and not the container in kubernetes deployments.

charts/keycloakx/values.yaml
```diff
@@ -236,7 +236,7 @@ service:
   # The HTTP Service node port if type is NodePort
   httpNodePort: null
   # The HTTPS Service port
-  httpsPort: 8443
+  httpsPort: null
```
grieshaber, line 239: Could you please revert the changes to the defaults? I agree to disable the https as a default would make sense. But, i do not have the time to "fix" the release-automation, so in order to merge this change sooner, i'd suggest to stick to the default for now.

grieshaber (changes requested):
It seems that the commit signature check is not succesful - https://github.com/codecentric/helm-charts/pull/651/checks?check_run_id=9886267385 could you have a look & commit with a valid signature?

charts/keycloakx/values.yaml
pyromaniac3010: done

PR #653, opened 2022-10-01
[keycloakx] Change default test image

cocor-au-lait:
There was a test image change in Keycloak.X helm upgrade.

https://github.com/codecentric/helm-charts/pull/633

I suggested using [joyzoursky/python-chromedriver](https://hub.docker.com/r/joyzoursky/python-chromedriver/) for this change, but it was pointed out that it is not currently actively maintained.

So this change proposes to use [seleniarm/standalone-chromium](https://hub.docker.com/r/seleniarm/standalone-chromium) as the test image instead.

This image is forked from selenium official image and actively maintained by the community, and supports multiple architectures.

Since the tag name corresponds to the chrome release version, the test target is clearer and it's possible to follow the next release quickly.

However, the main part of this image is selenium itself, so it doesn't include the `pip` and `selenium` module.

It seems that the mainstream method is to run two containers, one that runs headless browser via selenium and one running python test by connecting remote browser, but I didn't know how to do that flow on k8s.

So I added a setup script to the test flow.
Launch a headless browser and run test inside a single container as before.


grieshaber:
Thanks @cocor-au-lait for following up my comment in the other MR! Nice to use a more actively maintained image now.

I did dig any deeper by now, but on the first look, it seems that the lint and test charts job wasn't executed for this PR - any idea why?

Issue #652, opened 2022-10-01
Improve Versionbump for releases


grieshaber:
The current way of parsing commit messages and trying to detect the new version is not perfoming well, when semantic Commits are not used. This should be improved, so that Breaking changes are reflected in the version change aswell.

PR #664, opened 2022-11-02
[keycloakx] Bump Keycloak to v20.0.0

Footur:
I've updated the Keycloak container image to version v20.0.0.


grieshaber:
Thanks @Footur for keeping the keycloak image up2date. I think, we should quickly implement the Renovate Action for that to automate these kind of PRs.
Addtionally, before i want to merge this one here, we need to refactor the release mechanism (#652) do not release a major keycloak update as a minor / patch release again. This current setting was done by my predecessors and i did not completly dig into it, to understand how the release decisions are made.

ismailyenigul:
I would like to deploy keycloakx 20.X with helm chart.
Can we merge PR?

grieshaber (approved):
I pushed a commit with the "major"-keyword, this should trigger a major release on friday.

PR #665, opened 2022-11-06
[keycloakx] Quote all namespaces

dpunkturban:
Hi guys,

in this PR i quoted all namespaces so that you can install the helm chart in numeric only namespaces.

Without quoting:

```bash
helm upgrade --install --namespace 1234 keycloakx . --create-namespace

Release "keycloakx" does not exist. Installing it now.
Error: unable to build kubernetes objects from release manifest: unable to decode "": json: cannot unmarshal number into Go struct field ObjectMeta.metadata.namespace of type string
```


dpunkturban:
Hi,

any news from your side? Would be awesome to use the helm chart in numeric only namespaces :)

PR #669, opened 2022-11-24
Allow setting up a egress networkpolicy for Keycloak

wilmardo:
As stated in the title, allow the user to setup egress for Keycloak.

Useful when the database is external (within the cluster or outside the cluster) and the namespace has a default deny policy.

I choose to keep it broad `egress` to be able to specify any selector, the database can be in the cluster or outside the cluster. For example some networkplugins support a `DNSSelector`, other don't, etc.
All these subtle differences are allowed with the broader `egress` variable, the labels for keycloak are set, the rest is user input :)


grieshaber:
Hey @wilmardo - thanks for the PR. 

What do you think about adding the egress part to the existing network policy instead instanciating a complete new one?

That would make the setup less error prone, as it would only contain the metadata once.

wilmardo:
@grieshaber Thanks for the suggestion! Good improvement with the Egress in the same template.

I saw that the services where split over templates (without looking in the file :p) so I assumed that was a design preference and never re-evaluated that assumption.

grieshaber:
Hey @wilmardo - thanks for adapting. One more thing, before we can merge the egress addition: Could you please add the `networkPolicy.egress` to the Readme and some (minimal) information?
Than it should be completed imo :)

wilmardo:
I did? It is the top file change, I documented it like the `extraFrom` already was:

https://github.com/codecentric/helm-charts/pull/669/files#diff-eb66a25d719147755ab68fcda19c71871566c76a8654917eecc565a8193ff95eR153

grieshaber:
Oh right - i'm sorry. I checked something else and might be ended up in a wrong commit / branch..

grieshaber (approved):

Issue #670, opened 2022-12-05
Cannot set or update the MH_HOSTNAME var in mailhog chart


alxgomz:
# description:

[MH_HOSTNAME](https://github.com/mailhog/MailHog/blob/master/docs/CONFIG.md) is useful in order to tweak the `MAIL FROM` & `EHLO` commands  which are important in particular when releasing mails to real SMTP servers (which are picky about hostnames and email address to be real ones).
Unfortunately the current chart does not allow setting this variable. Or to be more precise, setting it result in an error when applying the template on upgrade:

```
│ Error: cannot patch "my-mailhog" with kind Deployment: Deployment.apps "my-mailhog" is invalid: spec.template.spec.containers[0].env[0].valueFrom: Invalid value: "": may not be specified when `value` is not empty
```

Currently the template always sets that variable to the name of the pod, which is problematic in most cases (as they do not map to real DNS names outside of the kubernetes cluster).

# how to reproduce:

do an initial install:

```sh
helm install my-mailhog --repo https://codecentric.github.io/helm-charts mailhog --version 5.2.1  -f -<<VALUES
extraEnv:
  - name: MH_HOSTNAME
    value: mail.mydomain.tld
VALUES
```
try to upgrade it

```sh
helm upgrade my-mailhog --repo https://codecentric.github.io/helm-charts mailhog --version 5.2.1 --reuse-values -f -<<VALUES
extraEnv:
  - name: MH_HOSTNAME
    value: newmail.mydomain.tld
VALUES
```

## additional notes:

Same issue if the MH_HOSTNAME var is added after the initial install

grieshaber:
PR is merged and will be released on friday afternoon

PR #671, opened 2022-12-05
feature: Allow setting MH_HOSTNAME for mailhog chart

alxgomz:
Treat MH_HOSTNAME as any other env var when set (otherwise keep previous behavior)
Aims at fixing #670


grieshaber (approved):
lgtm, thanks for PR

PR #678, opened 2023-01-08
comment ingress tls example to avoid conflicts

justmike1:
Hey!

I tried installing the helm chart with my values, I am advanced with helm, so I created my own values and ingress configuration, when I installed, I have noticed I encounter the error of:

```bash
Error syncing to GCP: error running load balancer syncing routine: error getting secrets for Ingress: secret "" does not exist
```

I never encountered this error so when googling I saw it is because I am using the GCE annotation in my ingress
```yaml
  annotations:
    networking.gke.io/managed-certificates: my-managed-certificate
    kubernetes.io/ingress.class: "gce"
```
and the tls configuration in the ingress:
```yaml
  tls:
    - hosts:
        - keycloak.example.com
      secretName: ""
```
got deployed to my ingress.

This PR will fix it as setting a tls config is not mandatory and there is other reasons to implement it to the ingress other than the forced implementation (like reverse proxies, etc.), I would appreciate it if this PR will be approved so I can clean my ingress config and as I am facing clients I would not want them to try figure it out the uanessacry troubleshooting it will cause.

Thanks,
Mike, a fellow DevOps Engineer. :)

Signed-off-by: Mike <90835468+justmike1@users.noreply.github.com>

### DIFF
--- a/charts/keycloak/values.yaml
+++ b/charts/keycloak/values.yaml
@@ -298,11 +298,11 @@ ingress:
       paths:
         - path: /
           pathType: Prefix
-  # TLS configuration
-  tls:
-    - hosts:
-        - keycloak.example.com
-      secretName: ""
+# Example TLS configuration
+#   tls:
+#     - hosts:
+#         - keycloak.example.com
+#       secretName: ""
 
   # ingress for console only (/auth/admin)
   console:


serdarkkts:
Hi @justmike1, I had a similar issue and fixed it by passing the `tls` section unset as follows. Hope that helps you and anyone who is dealing with the problem for now.

```yaml
# ingress configurations
ingress:
  enabled: true
  # leave empty for passing it unset, gcp managed certs do not need this.
  tls:
```

grieshaber (approved):
I understand the need of the change, thanks for contribution!

justmike1:
> Hi @justmike1, I had a similar issue and fixed it by passing the `tls` section unset as follows. Hope that helps you and anyone who is dealing with the problem for now.
> 
> ```yaml
> # ingress configurations
> ingress:
>   enabled: true
>   # leave empty for passing it unset, gcp managed certs do not need this.
>   tls:
> ```

@serdarkkts That's how I been using it til now, but now we don't need to.
Thanks!

justmike1:
@grieshaber btw, do you when the next chart bump?

PR #695, opened 2023-03-21
Update deprecated autoscaling/v2beta2 api

conor-nsurely:
The keycloakx chart is still using `autoscaling/v2beta2` for the HPA when autoscaling is enabled.

`autoscaling/v2beta2` has been deprecated since v1.23, and all k8s versions <= v1.23 are EOL.

I didn't make a PR to update the HPA API in the keycloak chart since it might make sense for that to remain on the legacy version. But I will make one if deemed appropriate.

### DIFF
--- a/charts/keycloakx/values.yaml
+++ b/charts/keycloakx/values.yaml
@@ -491,7 +491,7 @@ prometheusRule:
   #     severity: warning
 
 autoscaling:
-  # If `true`, a autoscaling/v2beta2 HorizontalPodAutoscaler resource is created (requires Kubernetes 1.18 or above)
+  # If `true`, an autoscaling/v2 HorizontalPodAutoscaler resource is created (requires Kubernetes 1.18 or above)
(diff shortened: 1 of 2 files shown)


charts/keycloakx/values.yaml
```diff
@@ -491,7 +491,7 @@ prometheusRule:
   #     severity: warning
 
 autoscaling:
-  # If `true`, a autoscaling/v2beta2 HorizontalPodAutoscaler resource is created (requires Kubernetes 1.18 or above)
+  # If `true`, an autoscaling/v2 HorizontalPodAutoscaler resource is created (requires Kubernetes 1.18 or above)
```
Boojapho, line 494:
```suggestion
  # If `true`, an autoscaling/v2 HorizontalPodAutoscaler resource is created (requires Kubernetes 1.23 or above)
```

grieshaber (approved):

PR #702, opened 2023-04-07
feat(keycloak): added modification of extraEnv as an object

justmike1:
Modifying extraEnv as one big string introduces issues overriding current values via terraform or helmfile with cherrypicking what to override instead to override the whole variable.

This feature fixes it

**fixed dco**

### DIFF
--- a/charts/keycloak/README.md
+++ b/charts/keycloak/README.md
@@ -239,9 +239,8 @@ Make sure you configure container support.
 This allows you to only configure memory using Kubernetes resources and the JVM will automatically adapt.
 
 ```yaml
-extraEnv: |
-  - name: JAVA_OPTS
-    value: >-
+extraEnv:
+  JAVA_OPTS: >-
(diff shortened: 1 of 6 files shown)


Boojapho:
@justmike1 The current implementation allows me to pass in references to secrets/configmaps to place into environmental variables.  But, the changes you added don't allow for that.  For example:

```yaml
extraEnv: |
  - name: KEYCLOAK_ADMIN_PASSWORD
    valueFrom:
      secretKeyRef:
        name: keycloak-admin
        key: password
```

I think your solution needs to account for this.  To avoid a breaking change, maybe you could just add another option called `extraEnvMap` that uses the key=value approach?

grieshaber:
Additionally, i'd like to separate the upgrade of postgres and the changes of the env-structure you made.

justmike1:
I see your point (both of you) and will change accordingly.

PR #708, opened 2023-04-25
feat(keycloakx): added extraenvmap value to pass

justmike1:
I would like the ability to pass map environment variables because I do not want to override default values and currently the `extraEnv` takes 1 big string as you always need to take default values together when you override.


justmike1:
I don't use old keycloak anymore, so closed the previous PR and opened updated one for keycloakx #702

justmike1:
@grieshaber can you please take a look at it and merge it if LGTY?

PR #716, opened 2023-06-16
feat(keycloak) Enable Keycloak cluster nodes discovery by default

oswaldo-montenegro-codecentric:
feat(keycloak) Include out-of-the-box high availability mechanism

While trying out the helm chart with replicas > 1, we've had the issue that
all keycloak instances don't identify each other to be part of the keycloak cluster.
This caused issues after logging in with multiple replicas. The problem lies in the missing distributed caching configuration. This change allows the helm chart to fully function with multiple replicas out of the box.

More information: https://www.keycloak.org/server/caching

Signed-off-by: Oswaldo Montenegro <oswaldo.montenegro@codecentric.de>

### DIFF
--- a/charts/keycloak/templates/statefulset.yaml
+++ b/charts/keycloak/templates/statefulset.yaml
@@ -101,6 +101,9 @@ spec:
                   name: {{ include "keycloak.postgresql.fullname" . }}
                   key: postgresql-password
             {{- end }}
+            {{- with .Values.jgroupsDiscovery }}
+            {{- tpl . $ | nindent 12 }}
+            {{- end }}
             {{- with .Values.extraEnv }}
             {{- tpl . $ | nindent 12 }}
             {{- end }}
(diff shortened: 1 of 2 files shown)


devRoemer:
I think that having the env vars added depending on replica number is not flexible enough.
There might be use cases where one wants to add those vars via extraEnv with different values.

For this case, there should at least be a boolean in values.yaml to disable automatically adding them (if not even the entire config of both values).

oswaldo-montenegro-codecentric:
@devRoemer Thanks for your comment! I refactored my change in a way that is more flexible to change and adapt.

grieshaber:
Hi @oswaldo-montenegro-codecentric, thanks for the PR. Imho the defaults seem to be mixed right now.. with the default values, keycloak is installed with a single replica and autoscaling disabled, but your changes imply default cluster nodes discovery mechanisms.

From my pov, we should comment your changes, so that they are included in the values for the reference but disabled as a default. I'm not in the topic enough to verify, whether your changes do have negative impact in non-cluster mode.

PR #744, opened 2024-01-24
Remove Keycloak JDK UseContainerSupport parameter

megian:
The parameter `-XX:+UseContainerSupport` is no longer required for Keycloak >= v21 based on JDK v17. JDK >= v17 has enabled the container support by default.


megian:
@pascal-sochacki Might you be able to review + merge this?

PR #762, opened 2024-04-23
feat(keycloakx): Support optimized start

mruzicka:
This PR makes the chart aware of the so called _optimized_ start of Keycloak. This optimized start is achieved by passing the `--optimized` argument to the Keycloak's `start` command:
```
kc.[sh|bat] start --optimized ...
```
When Keycloak is started this way it assumes it had already been optimized (see [Creating an optimized Keycloak build](https://www.keycloak.org/server/configuration#_creating_an_optimized_keycloak_build)) and ignores options which only take effect during the optimization process but does log a warning when it encounters them. The warning may look something like this:
```
WARN  [org.keycloak.quarkus.runtime.cli.Picocli] (main) The following build time non-cli options were found, but will be ignored during run time: kc.cache, kc.cache-stack, kc.http-relative-path, kc.health-enabled, kc.metrics-enable
```
The intent of this PR is to get rid of this warning by not passing the build/optimization-time only options when an optimized start is used. For that purpose a boolean value `optimized` is introduced which is used to guard the rendering of the build-time only options.

Note that the `kc.cache*` options are in fact run-time options (even though mentioned in the example warning above). This is a bug in the definition of these options in Keycloak. It's already been fixed in https://github.com/keycloak/keycloak/commit/a3669a6562dd190b2ec48e10de8250a3fb7fcfca, so once this is released, Keycloak will stop complaining about those.


mruzicka:
I've updated the PR, could you please reopen it?

PR #770, opened 2024-06-14
Upgraded to version 25.0.0 of Keycloak

hansehe:
Upgraded to version 25.0.0 of Keycloak. Tested in separate systems, and only change needed was to use the new internal 9000 port for health checks and metrics.

### DIFF
--- a/charts/keycloakx/templates/statefulset.yaml
+++ b/charts/keycloakx/templates/statefulset.yaml
@@ -100,8 +100,12 @@ spec:
               value: "kubernetes"
             {{- end }}
             {{- if .Values.proxy.enabled }}
-            - name: KC_PROXY
+            - name: KC_PROXY_HEADERS
               value: {{ .Values.proxy.mode }}
+            {{- end }}
+            {{- if .Values.proxy.http.enabled }}
+            - name: KC_HTTP_ENABLED
+              value: "true"
             {{- end }}
             {{- if .Values.database.vendor }}
             - name: KC_DB
--- a/charts/keycloakx/templates/statefulset.yaml
+++ b/charts/keycloakx/templates/statefulset.yaml
@@ -145,6 +145,9 @@ spec:
             - name: http
               containerPort: 8080
               protocol: TCP
+            - name: http-internal
+              containerPort: 9000
+              protocol: TCP
             {{- if .Values.service.httpsPort }}
             - name: https
               containerPort: 8443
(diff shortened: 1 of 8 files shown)


amarkevich:
Can be internal port parametrized to be compatible with previous Keycloak versions? (just an idea)

hansehe:
Yep, sounds good, and think my latest commit should fix it. I kept the http-internal port there for all cases, since it shouldn't matter if an extra port is exposed, although not used.

hansehe:
Seems like the lint tests failed during the test setup.

pascal-sochacki:
Hey there, I just received access to this Repo and will look at the failing Pipeline

Theoshen (approved):

pascal-sochacki:
Hey, I updated the failing job to the newest version in #772, could you rebase your branch, and we will see if it works now?

hansehe:
Yes, rebased now and it seems to work, and thanks.

charts/keycloakx/templates/statefulset.yaml
```diff
@@ -145,6 +145,9 @@ spec:
             - name: http
               containerPort: 8080
               protocol: TCP
+            - name: http-internal
```
pascal-sochacki, line 148:
```suggestion
            - name: '{{ .Values.http.internalPort }}'
```

pascal-sochacki (commented):
Can you look at the Pipeline? It look like further adjustments are required for the new Version.

charts/keycloakx/templates/statefulset.yaml
hansehe: Fixed it

hansehe: Anything else missing?

hansehe: Seems lik I missed removing a deprecated argument, this the --hostname-strict-https argument. It should be working now.

hansehe: After testing, I also discovered the deprecated proxy option. So I updated the values from the upgrade documentation here:
- https://www.keycloak.org/docs/latest/upgrading/index.html#deprecated-proxy-option

I think it will be too difficult to try a full backwards compatibility version for this one. If someone wants to run an older version of Keycloak, then I would advice turning the proxy of in the values file and adding necessary env variables manually.

pascal-sochacki: Looks good. Just sign off your commit and I will merge it and figure out how to release this 😅 
https://github.com/codecentric/helm-charts/pull/770/checks?check_run_id=26551409269

hansehe: Signed and done

karazy:
Great to see support for v25. Thanks a lot. Please create a release.

PR #782, opened 2024-07-12
FIX: rendering of the internal port in the serviceMonitor

binlan:
fix the rendering of the internal port in the serviceMonitor


igor-nikiforov:
@pascal-sochacki could you please review this PR? Thanks!

Issue #781, opened 2024-07-11
Keycloakx 25.0.1 crashloop without any message


beezerk23:
Hey, im also running into a crashloop. For whatever reason, all my deployment is doing is this:

```
Appending additional Java properties to JAVA_OPTS
Updating the configuration and installing your custom providers, if any. Please wait.
2024-07-11 12:38:47,833 INFO  [io.qua.dep.QuarkusAugmentor] (main) Quarkus augmentation completed in 110325ms
2024-07-11 12:39:44,173 INFO  [org.infinispan.CONTAINER] (Thread-5) ISPN000556: Starting user marshaller 'org.infinispan.jboss.marshalling.core.JBossUserMarshaller'
2024-07-11 12:39:48,152 INFO  [org.infinispan.CLUSTER] (Thread-5) ISPN000078: Starting JGroups channel `ISPN` with stack `kubernetes`
2024-07-11 12:39:48,162 INFO  [org.jgroups.JChannel] (Thread-5) local_addr: 80abec56-ff79-4a91-b44d-2799b0a24108, name: keycloak-keycloakx-0-61709
2024-07-11 12:39:48,265 INFO  [org.jgroups.protocols.FD_SOCK2] (Thread-5) server listening on *.57800
2024-07-11 12:39:50,311 INFO  [org.jgroups.protocols.pbcast.GMS] (Thread-5) keycloak-keycloakx-0-61709: no members discovered after 2008 ms: creating cluster as coordinator
2024-07-11 12:39:50,690 INFO  [org.infinispan.CLUSTER] (Thread-5) ISPN000094: Received new cluster view for channel ISPN: [keycloak-keycloakx-0-61709|0] (1) [keycloak-keycloakx-0-61709]
2024-07-11 12:39:54,320 INFO  [org.infinispan.CLUSTER] (Thread-5) ISPN000079: Channel `ISPN` local address is `keycloak-keycloakx-0-61709`, physical addresses are `[10.8.2.106:7800]`
2024-07-11 12:39:57,583 INFO  [org.keycloak.broker.provider.AbstractIdentityProviderMapper] (main) Registering class org.keycloak.broker.provider.mappersync.ConfigSyncEventListener
2024-07-11 12:39:59,916 INFO  [org.keycloak.connections.infinispan.DefaultInfinispanConnectionProviderFactory] (main) Node name: keycloak-keycloakx-0-61709, Site name: null
[...]

PR #784, opened 2024-07-30
feature: Set scheme for probes and serviceMonitor

### LINKED ISSUE #778: Keycloak 25.0.0 Helm Chart deployment -> CrashLoop
Upgraded keycloak from 22.0..4 to 25.0.0 using new keycloakx chart as a drop in replacement - same helm cmd line, same values.yaml. The service starts and fails/CrashLoops. 
Initial error:
Startup probe failed: Get "http://10.42.2.163:9000/auth/health": dial tcp 10.42.2.163:9000: connect: connection refused

I disabled health check by adding health.enabled=false to the values.yaml, but it still crashes. Nothing in the service logs besides this:

`Appending additional Java properties to JAVA_OPTS
Changes detected in configuration. Updating the server image.
Updating the configuration and installing your custom providers, if any. Please wait.
2024-06-29 00:19:35,792 INFO  [io.qua.dep.QuarkusAugmentor] (main) Quarkus augmentation completed in 16683ms
Server configuration updated and persisted. Run the following command to review the configuration:
	kc.sh show-config
Next time you run the server, just run:
	kc.sh --verbose start --http-enabled=true --http-port=8080 --hostname-strict=false --hostname-strict-https=false --spi-events-listener-jboss-logging-success-level=info --spi-events-listener-jboss-logging-error-level=warn --optimized
Disabled option: '--hostname-strict-https'. Available only when hostname:v1 feature is enabled
Possible solutions: --hostname, --hostname-admin, --hostname-backchannel-dynamic, --hostname-strict, --hostname-debug
Try 'kc.sh start --help' for more information on the available options.
Specify '--help-all' to obtain information on all options and their availability

DB checker:

`Waiting for Database to become ready...
.Database OK ✓`

CrashLoopBackOff: back-off 5m0s restarting failed container=keycloak

TheBigLee:
The new health endpoints used by the three probes (liveness, readiness
and startup) as well as the metrics endpoint are now served on a
dedicated interface with its own port. The scheme of the interface only
servers HTTP or HTTPS. If the main keycloak interface serves HTTPS,
the internal management interface will also serve HTTPS, regardless if
the main keycloak interface servers plain HTTP or not.

It's therefore impossible for the probes or the serviceMonitor to work
with a plain HTTP scheme, if keycloak has HTTPS enabled.

We therefore need a way to specify the scheme of the internalPort via
helm values to be able to set a custom scheme for the probes and the
serviceMonitor

This fixes #778 and #781


mruzicka:
@TheBigLee , the scheme should not be hardcoded, it would break the probes for everyone (us included) who runs Keycloak on http (behind a SSL terminating proxy) by providing these values to the chart:
```yaml
proxy:
  enabled: true
  mode: xforwarded
  http:
    enabled: true
```
I.e. the scheme should be conditional on the `.Values.enabled` _and_ `.Values.http.enabled` values.

TheBigLee:
> @TheBigLee , the scheme should not be hardcoded, it would break the probes for everyone (us included) who runs Keycloak on http (behind a SSL terminating proxy) by providing these values to the chart:
> 
> ```yaml
> proxy:
>   enabled: true
>   mode: xforwarded
>   http:
>     enabled: true
> ```
> 
> I.e. the scheme should be conditional on the ~`.Values.enabled` _and_~ `.Values.http.enabled` value~s~.
> 
> Edit: It turns out it is enough to set `.Values.http.enabled` to `true`, for Keycloak to be run on http.

It's actually more complicated than that.
Keycloak itself can run on http, https or both. The management interface however, can only run on http or https. If Keycloak itself is running on https than the management interface will run on https as well.
The chart itself currently doesn't support setting up HTTPS directly for Keycloak. You have to set the `KC_HTTPS_CERTIFICATE_FILE` env variable via the  `extraEnv` chart variable and load the certificate and key using the `extraVolumes` chart variable.

So there is actually no easy way to automatically determine the correct scheme for the probes.
I've refactored the code and defined a new variable `http.internalScheme` that can easily be overridden in the chart. Hopefully this approach is acceptable.

Issue #796, opened 2024-09-18
Keycloak 25 / keycloakx - ServiceMonitor does not work with default port property


maximilian-krauss:
Heyho,
after the upgrade to Keycloak 25 and Chartversion 2.5.1 we are not getting any metrics via the Prometheus Operator.

We noticed that since Keycloak 25 metrics are being exposed only on Port 9000. We had to change `port` in the ServiceMonitor to `targetPort` in order to make it work. 

Unfortunately it is not possible to overwrite this property in this chart. But before sending a PullRequest to change it here I want to ask if someone is able to pull the metrics with the default chart configuration?

Tim-Schwalbe:
No, I just have the same problem. I don't get any metrics and not the UP metric

maximilian-krauss:
Hah, if it's not just us, then I will send a PR to fix it 🙂 

I changed it directly via kubectl in the ServiceMonitor after the helm deployment.

You need to replace `port` with `targetPort` and then it should work after a few minutes.

Tim-Schwalbe:
It not working for me this way. I get the Up Metrics now, but somehow it's always 0, but keycloak started and all seems to work.

Tim-Schwalbe:
okay I think I see the problem here. The service is not even listening to port http-internal aka 9000. So I had to extend the service-http.yaml 

but I am really asking myself how it could work for you @maximilian-krauss 

`  ports:
    - name: '{{ .Values.http.internalPort }}'
      port: 9000
      protocol: TCP
      targetPort: '{{ .Values.http.internalPort }}' `

maximilian-krauss:
That's weird. Does it work with the port property on the service monitor after you added the port to the service? 

As far as I understood that's the difference between port and targetPort:
![CleanShot 2024-09-20 at 12 44 48@2x](https://github.com/user-attachments/assets/ebb98aea-a021-498b-b363-15e2de03b9b9)
from: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#endpoint

By using targetPort in the service monitor it uses the pod to get metrics, port tries to use the service. 

I will also try on my setup which combination works, if it's the service its easy to add via the 'extraPorts' property

Tim-Schwalbe:
Yes once I added the port to the service I worked. I am just tried with targetPort and it's also working :D
So both is fine I think I will go with the targetPort Version. Looks cleaner for me. I think I had also to fix my networkpolicies that's why it was not work in the first place.

[...]

PR #798, opened 2024-09-20
[keycloakx/keycloak 25] add internalPort to service-http.yaml

Tim-Schwalbe:
#796 

To be able to get the metrics including the UP metric it is necessary to add the metrics endpoint/port to the service itself.

### DIFF
--- a/charts/keycloakx/templates/service-http.yaml
+++ b/charts/keycloakx/templates/service-http.yaml
@@ -35,6 +35,10 @@ spec:
   {{- end }}
   {{- end }}
   ports:
+    - name: '{{ .Values.http.internalPort }}'
+      port: 9000
+      protocol: TCP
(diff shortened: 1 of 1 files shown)


wilmardo:
Pretty sure this PR is still relevant, I had to add this to the values to get the metrics to work:

```yaml
          extraPorts:
            - name: http-internal
              port: 9000
              targetPort: http-internal
              protocol: TCP
```

It would be nice if this could be merged so it works like it did before the 9000 portchange in Keycloak upstream.

maximilian-krauss:
Hi @wilmardo,
yes it definitely is still relevant.

@Tim-Schwalbe could you reopen the PR and add the signed-off header to the commit that someone from the maintainers can merge it?

Cheers,
Max

Issue #797, opened 2024-09-19
Provide charts through OCI registry


jonesbusy:
Provide OCI chart in addition of classical registry

The easiest is to use GItHub package (ghcr.io) to publish OCI chart packages when they are released

Similar works as been done on many chart repositories (https://github.com/grafana/helm-charts/issues/2395, https://github.com/jenkinsci/helm-charts/pull/1196/files, etc...)

Can we update the release workflow to also publish OCI helm charts ?

I volonteer to also open a PR

PR #799, opened 2024-09-20
feature: publish to GHCR registry

jonesbusy:
Fix : https://github.com/codecentric/helm-charts/issues/797

Charts will now be published in addition to traditional helm deployment also in GitHub container registry to be consumed by OCI client or directly since Helm 3.8

I've tested the release on my fork

release_process1

This is similar to other OpenSource charts: https://github.com/jenkinsci/helm-charts/pull/1196 or https://github.com/sigstore/helm-charts/pull/43

Hope this contribution will be accepted so that we can proxy/pull such OCI artifact

### DIFF
--- a/.github/workflows/release-charts.yaml
+++ b/.github/workflows/release-charts.yaml
@@ -76,7 +76,7 @@ jobs:
       - name: Install Helm
         uses: azure/setup-helm@v1
         with:
-          version: v3.16.1
+          version: v3.16.3
 
       - name: Add dependency chart repos
         run: |
@@ -102,5 +102,5 @@ jobs:
             if [ -z "${pkg:-}" ]; then
               break
             fi
-            helm push "${pkg}" oci://ghcr.io/jonesbusy/helm-charts
+            helm push "${pkg}" oci://ghcr.io/codecentric/helm-charts
           done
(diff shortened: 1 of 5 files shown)


jonesbusy:
Hi @pascal-sochacki What do you think about this PR ? Do you think `codecentric` can also be published as OCI artifact on this repository ? Thanks!

.github/workflows/release-charts.yaml
```diff
@@ -71,15 +76,31 @@ jobs:
...
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"          
+
+      - name: Push Chart to GHCR
+        run: |
+          for pkg in .cr-release-packages/*; do
+            if [ -z "${pkg:-}" ]; then
+              break
+            fi
+            helm push "${pkg}" oci://ghcr.io/jonesbusy/helm-charts
```
pascal-sochacki, line 105: I think you need to use variables, so forks and this repository work

jonesbusy: Good catch. I used it for testing and forgot to put it back

.github/workflows/release-charts.yaml
```diff
@@ -102,5 +102,5 @@ jobs:
             if [ -z "${pkg:-}" ]; then
               break
             fi
-            helm push "${pkg}" oci://ghcr.io/jonesbusy/helm-charts
+            helm push "${pkg}" oci://ghcr.io/codecentric/helm-charts
```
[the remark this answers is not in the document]
jonesbusy: Sure fine for me

Issue #802, opened 2024-10-29
Truncate app.kubernetes.io/version label


sylvainmetayer:
If using a Keycloak image fixed with a sha256, the label is too long (for example `25.0.6@sha256:3efa9e1f14539d138a86c0169df805f692bbc984a43ceb15bb04074fec8b1986`).

The label should be truncated at 63 characters to ensure it won't produce an error.

PR #803, opened 2024-10-29
fix: add truncate filter on version label

sylvainmetayer:
Fix https://github.com/codecentric/helm-charts/issues/802


pascal-sochacki:
Could you rebase, please? The failing Job should be fixed now

PR #808, opened 2024-12-04
feature: Allow overriding the backend name and port in keycloakx chart

hpdvanwyk:
This is useful for overriding the backend of certain paths to not go to Keycloak. 
One use case is when using Keycloak with the https://github.com/aerogear/keycloak-metrics-spi plugin along with aws-load-balancer-controller. This allow using the "alb.ingress.kubernetes.io/actions.${action-name}" annotation to make in-realm metrics endpoints 404 instead of exposing them on the public internet. This would otherwise be annoying to achieve since this plugin puts metrics on `<base url>/realms/<realm>/metrics`.

For example this allows creating the following ingress object which will 404 on all <base url>/realms/<realm>/metrics paths while allowing everything else under <base url>/realms/:
```
kind: Ingress
metadata:
  name: release-name-keycloakx
  namespace: default
  annotations:
    alb.ingress.kubernetes.io/actions.response-metrics: "{\"type\":\"fixed-response\",\"fixedResponseConfig\":{\"statusCode\":\"404\"}}\n"
spec:
  ingressClassName: alb
  tls:
    - hosts:
        - "${host}"
  rules:
    - host: "${host}"
      http:
        paths:
          - path: "/auth/realms/*/metrics"
            pathType: ImplementationSpecific
            backend:
              service:
                name: response-metrics
                port:
                  name: use-annotation
          - path: "/auth/realms/*"
            pathType: ImplementationSpecific
            backend:
              service:
                name: release-name-keycloakx-http
                port:
                  name: http
```
with the values.yaml snippet:
```
  rules:
    - host: "${host}"
      paths:
        - path: '{{ tpl .Values.http.relativePath $ | trimSuffix "/" }}/realms/*/metrics'
          pathType: ImplementationSpecific
          backendOverride:
            name: response-metrics
            port: use-annotation
        - path: '{{ tpl .Values.http.relativePath $ | trimSuffix "/" }}/realms/*'
          pathType: ImplementationSpecific
```


stefangries:
Thank you for your contribution. Please adjust the naming of the variables so that they match the naming conventions of this chart.

Features are switched on via .enabled, for example. Please also add the default value ‘false’, so that we remain backward compatible.

stefangries:
@hpdvanwyk Still interested in getting this merged? Please see my previous comment :-)

hpdvanwyk:
We went with a completely different approach to keep metrics endpoints private so there is no real incentive for me to try to get this merged any more.

PR #822, opened 2025-03-05
fix: toString before truncate

### LINKED ISSUE #821: [keycloak] [keycloakx] Image tag is truncated like string but can be also number
When image tag is just a number (which can be since #640) it fails due truncating done in #802.
```
Error: UPGRADE FAILED: template: keycloakx/templates/statefulset.yaml:13:8: executing "keycloakx/templates/statefulset.yaml" at <include "keycloak.labels" .>: error calling include: template: keycloakx/templates/_helpers.tpl:38:84: executing "keycloak.labels" at <63>: wrong type for value; expected string; got int64
```

We should make sure the version is string before truncating.
```
app.kubernetes.io/version: {{ .Values.image.tag | default .Chart.AppVersion | toString | trunc 63 | quote }}
```

okosatka:
This commit prevents failing deployment when `.Values.image.tag` (in our case shortened commit hash) is just a number (like `4524551` - contains just decimal digits, without any `a`-`f`)
Resolves #821

### DIFF
--- a/charts/keycloak/templates/_helpers.tpl
+++ b/charts/keycloak/templates/_helpers.tpl
@@ -37,7 +37,7 @@ Common labels
 {{- define "keycloak.labels" -}}
 helm.sh/chart: {{ include "keycloak.chart" . }}
 {{ include "keycloak.selectorLabels" . }}
-app.kubernetes.io/version: {{ .Values.image.tag | default .Chart.AppVersion | trunc 63 | quote }}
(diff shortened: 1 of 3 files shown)


okosatka:
Could you please review my pull request when you have a moment? I’d appreciate your feedback. Thanks!
cc: @stefangries

okosatka:
reopen

PR #827, opened 2025-04-22
[keycloakX] add tlsConfig to serviceMonitor

### LINKED ISSUE #825: [keycloakx] servicemonitor tlsConfig
Hi,
we use self-signed certificates for the keycloak pods in the backend and then _real_ certificates in the Ingress Controller.
When configuring the Prometheus Operator, I need to set `tlsConfig.insecureSkipVerify=true` so prometheus does not check the certificates. 
I can prepare a PR to add this functionality.

belfhi:
fixes #825


belfhi:
yes, sorry I rebased my branch and force pushed 👍

stefangries:
Thanks!

sahil-sharma:
I think ``tlsConfig`` is not being released on 7.1.1 version. Correct me if I am wrong?
As I can't see it in its [YAML](https://github.com/codecentric/helm-charts/blob/keycloakx-7.1.1/charts/keycloakx/templates/servicemonitor.yaml#L34-L47).

stefangries:
It will be released in the next version in a few days.

PR #831, opened 2025-05-12
change ServiceAccount RBAC to Role & ClusterRole

### LINKED ISSUE #829: [keycloakx] Remove ServiceAccount Cluster Permissions
Currently, the helm chart includes RBAC for `ClusterRole` and `ClusterRoleBinding` for the serviceAccount.
This is not neccessary since the keycloak helm release is in a single namespacs.
Thus, a `Role` and `RoleBinding` should be sufficient to use all the DNS lookup of pod names etc  in the StatefulSet.

belfhi:
ClusterRole and ClusterRoleBinding is not needed, use Role and RoleBinding instead, fixes #829


stefangries:
@belfhi Do you think we should mark this as a breaking change? I'm currently thinking about various scenarios for the update and the effects of this change on existing installations.

PR #832, opened 2025-05-19
[mailhog] add commonLabels to chart

ashleydavies:
Adds `commonLabels` so that you can add labels to all resources easily, useful for e.g. if you need to comply with an admission policy that requires ownership labels.


stefangries:
Please also add and entry to the Readme File and add a commented out example to the values file.

PR #834, opened 2025-06-03
Enable RelabelConfigs to apply to samples before scraping

mehr74:
The goal of this PR is to enable adding and updating labels in serviceMonitor. The features is needed, when we have multiple environments and we would like to enable specific labels for instance of Keycloak in each environment.


stefangries:
You added this to the very old keycloak chart, which currently is von Keycloak 17.0.1. Is that really what you wanted to do?
The current keycloakx chart already supports this feature and has the current keycloak version.

PR #841, opened 2025-07-30
feat: add separate labels to console ingress

andrewjkeith:
add the option to separately label the console ingress


stefangries:
Doesn't that have to be added to values.schema.json as well?

jseiser:
@andrewjkeith Any update on getting the values updated on this one?

andrewjkeith:
@stefangries i went ahead at refactored the values.schema.json, turns out it had several errors in there. let me know if you want a separate pull request

stefangries:
Yes, please do separate pull requests for the refactoring and for this issue.

andrewjkeith:
@stefangries ok i just opened a new PR here, thanks: https://github.com/codecentric/helm-charts/pull/858

stefangries:
Thank you. If I'm not mistaken, the changes to values.schema.json are currently in both this pull request #841 and the other one #858. Could you please remove them from this pull request so that they are not in both?

Please excuse the late reply.

andrewjkeith:
> @andrewjkeith First, please excuse my late reply. I have now looked at your pull request again. As I see it, this is a breaking change, as labels that may have previously been under $ingress.labels are no longer added. The new naming is more correct, but it is also a breaking change. Could we perhaps achieve this without a breaking change by adding both fields, $ingress.labels and $ingress.console.labels? What do you think?

yeah that is a good point, i agree. i added back the `ingress.labels`

Issue #835, opened 2025-07-04
[keycloakx] Missing JGroups port in headless service


nianlongyu-geotab:
The `service-headless.yaml` template is missing the JGroups peer discovery port `7800`. When Istio service mesh is used in the environment, this missing port definition in the service object will cause Istio sidecar proxy to drop all outbound traffic sent by Keycloak to its peers on this port.

stefangries:
Hi @nianlongyu-geotab! Thanks for reporting this. Would you like to create a fix and merge request?

stefangries:
Thank you. Please also add this to the Readme.

PR #850, opened 2025-08-21
[keycloakx] hotfix on incorrect templating of serviceHeadless.extraPorts

nianlongyu-geotab:
Hotfix on the new feature introduced in #847, which originates from #835.

Replaced incorrect use of the `range` template with `with` template.


nianlongyu-geotab:
@stefangries could you please provide the review on this PR that addresses a mistake I made in my previous contribution?

PR #869, opened 2025-11-20
feat: Add overrides for ingress backend service name and port

elkh510:
This change adds support for overriding the ingress backend service name and port via the serviceName and servicePort values while keeping the existing defaults.

These overrides are required to support AWS ALB “action” annotations such as:

```
alb.ingress.kubernetes.io/actions.block-master-account: |
  {
    "type": "fixed-response",
    "fixedResponseConfig": {
      "contentType": "text/plain",
      "statusCode": "403",
      "messageBody": "Access Forbidden"
    }
  }

alb.ingress.kubernetes.io/actions.redirect-root: |
  {
    "type": "redirect",
    "redirectConfig": {
      "statusCode": "HTTP_301",
      "path": "/realms/shared/account",
      "query": "#{query}"
    }
  }
```

In these cases, the ingress backend must reference a logical action name instead of a real Kubernetes Service:

```
paths:
  - path: /realms/master/account/
    pathType: Prefix
    backend:
      service:
        name: block-master-account
        port:
          name: use-annotation
```

Without the ability to override serviceName and servicePort, such ALB action backends cannot be expressed through the chart values.

### DIFF
--- a/charts/keycloakx/Chart.yaml
+++ b/charts/keycloakx/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: keycloakx
-version: 7.1.5
-appVersion: 26.4.5
+version: 7.1.7
+appVersion: 26.5.1
 description: Keycloak.X - Open Source Identity and Access Management for Modern Applications and Services
 keywords:
   - sso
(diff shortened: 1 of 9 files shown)


elkh510:
@jenslehnhoff hi
added changes
pls take a look

PR #870, opened 2025-11-20
feat: Support custom manifests through the extraManifests value

elkh510:
This change introduces the extraManifests value, allowing users to provide and render arbitrary additional Kubernetes manifests as part of the chart deployment.
This enables injecting custom resources (such as ConfigMaps or any other Kubernetes objects) directly through values without modifying the chart itself.

### DIFF
--- a/charts/keycloakx/README.md
+++ b/charts/keycloakx/README.md
@@ -593,6 +593,32 @@ annotations:
     }
 ```
 
+### Extra Kubernetes Manifests
+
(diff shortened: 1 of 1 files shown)


elkh510:
@stefangries hi
Could you spare a few minutes for this PR?

jperville:
Interested in this PR, any idea if a merge could come soon?

I want to add an kubernetes Job that add a new keycloak admin and remove the bootstrap admin using kcadm.sh in a toy environment without having to wrap the keycloakx chart.

PR #871, opened 2025-11-21
feat: adds gateway api httproute support

wilmardo:
Adds support for the HTTPRoute resource.

I adapted the default template/values that came with `helm create chart` to have it uniform with other helm charts. I kept it as close to the current ingress setup as possible by splitting console and allowing templating on the path.

The templating on the path meant however that (for now) only the path matcher is supported since I needed to traverse the dict to be able to run `tpl`. This is represented in the schema, it won't allow other matchers then path.

I think this is the best solution to keep backwards compatible and adding other matcher is still possible, you will just need to add another clause to the if. Kept it simple for now.

### DIFF
--- a/charts/keycloakx/templates/httproute.yaml
+++ b/charts/keycloakx/templates/httproute.yaml
@@ -4,6 +4,7 @@ apiVersion: gateway.networking.k8s.io/v1
 kind: HTTPRoute
 metadata:
   name: {{ include "keycloak.fullname" . }}
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "keycloak.labels" . | nindent 4 }}
     {{- range $key, $value := $httpRoute.labels }}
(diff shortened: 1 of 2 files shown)


Eifoen:
Is this ready to be merged? Looks pretty solid to me! Would be great to have this soon, as a lot of people are currently switching from nginx-ingress and are looking into switching ingress->gateway-api. 😊

jenslehnhoff (changes requested):
Hi @wilmardo 

Thank you for your changes. I apologize for the delayed response.

We should set the HTTPRoute to false by default and add the namespace as a best practice.

Could you also add the parameters for the HTTPRoute feature to the README, please?

After these changes, we can finally merge your pull request 🥳

Regards
Jens

rohit-dimagi:
Any idea when this will be released?

PR #887, opened 2026-03-11
Add compatibility with OKD and OpenShift

hlavki:
This change adds compatibility with **OKD** and **OpenShift**.

The rendering of security contexts has been adjusted to allow the container to run on these platforms without explicitly setting UID and GID. This aligns with the security model used by OKD/OpenShift, where these values are typically assigned dynamically.

The implementation is based on logic from the Bitnami Helm charts repository (Apache License).

The change preserves backward compatibility with existing configurations.


charts/keycloakx/templates/statefulset.yaml
```diff
@@ -46,8 +46,9 @@ spec:
         - name: dbchecker
           image: "{{ .Values.dbchecker.image.repository }}{{- if (.Values.dbchecker.image.digest) -}}@{{ .Values.dbchecker.image.digest }}{{- else -}}:{{ .Values.dbchecker.image.tag }} {{- end }}"
           imagePullPolicy: {{ .Values.dbchecker.image.pullPolicy }}
-          securityContext:
-            {{- toYaml .Values.dbchecker.securityContext | nindent 12 }}
+          {{- if .Values.podSecurityContext.enabled }}
+          securityContext: {{- include "keycloak.renderSecurityContext" (dict "secContext" .Values.podSecurityContext "context" $) | nindent 12 }}
```
dominiquemetz, line 50: This security context belongs to the dbchecker, which should be set via .Values.dbchecker.securityContext. Please adapt your solution to use this securityContext

charts/keycloakx/values.yaml
```diff
@@ -68,12 +68,22 @@ rbac:
...
+  ## Compatibility adaptations for Openshift
+  ##
+  openshift:
+    ## @param compatibility.openshift.adaptSecurityContext Adapt the securityContext sections of the deployment to make them compatible with Openshift restricted-v2 SCC: remove runAsUser, runAsGroup and fsGroup and let the platform use their allowed default IDs. Possible values: auto (apply if the detected running cluster is Openshift), force (perform the adaptation always), disabled (do not perform adaptation)
+    ##
+    adaptSecurityContext: auto
+
 # SecurityContext for the entire Pod. Every container running in the Pod will inherit this SecurityContext. This might be relevant when other components of the environment inject additional containers into running Pods (service meshes are the most prominent example for this)
 podSecurityContext:
+  enabled: true
```
dominiquemetz, line 81: Are the "enabled" flags necessary for this feature? Why would I need to disable the rendering of the security contexts in the context of OKD and OpenShift support? If not strictly necessary, I would ask you to remove the enabled flag and have no conditional rendering of the securityContexts

dominiquemetz (changes requested):
Thank your for your contribution, this already looks great!
I added two comments, please take a look.

Please update the Configuration section in the README and document the adaptSecurityContext property as well.

PR #898, opened 2026-04-25
Keycloakx: Automate minor/major upgrade process in cluster mode

maarous:
Keycloak does **not** support minor or major version upgrades while running in cluster mode. Only patch updates are supported.
(Refer to the official Keycloak documentation for more details.)

If you attempt such an upgrade in cluster mode, you may encounter errors due to JGroups version mismatches, for example:

> WARN [org.jgroups.protocols.TCP] (TcpServer.Acceptor[7800]-1,keycloakx-1-2180(v=16.0.8)) JGRP000006: 10.151.254.47:7800: failed accepting connection from peer SSLSocket[hostname=127.0.0.6, port=51749, Session(1776862509121|TLS_AES_256_GCM_SHA384)]: java.io.IOException: 10.151.254.47:7800: readPeerAddress(): packet from /127.0.0.6:51749 has different version (5.3.16) from ours (5.5.1); discarding it

Because of this limitation, **downtime is required** to perform minor or major version upgrades.

**Recommended upgrade procedure:**

1. Scale down the StatefulSet to a single replica to disable cluster mode.
2. Perform the version upgrade.
3. Once the upgrade is complete, scale the StatefulSet back to its original number of replicas.

To simplify this process, this pull request introduces a new parameter `updateHook` that automates these steps.

You can see in this GIF a test of upgrading a three-instance cluster from version 26.5.6 to version 26.6.1.

Screencast from 2026-04-25 21-37-42

---

I took this opportunity to upgrade Keycloak to the latest version, 26.6.1.


dominiquemetz:
Hi! When trying out your feature, my local cluster of 3 Keycloaks scales down, but not up anymore.

> `helm upgrade --install keycloak helm-charts/charts/keycloakx --values values.yaml --namespace pr898 --set image.tag=26.6.1`
`level=WARN msg="upgrade failed" name=keycloak error="conflict occurred while applying object pr898/keycloak-keycloakx apps/v1, Kind=StatefulSet: Apply failed with 1 conflict: conflict with \"curl\" with subresource \"scale\" using apps/v1: .spec.replicas"`
`Error: UPGRADE FAILED: conflict occurred while applying object pr898/keycloak-keycloakx apps/v1, Kind=StatefulSet: Apply failed with 1 conflict: conflict with "curl" with subresource "scale" using apps/v1: .spec.replicas
`

The `curl` command modifies and takes ownership of the `scale` resource:

> ... 
>  apiVersion: apps/v1
    fieldsType: FieldsV1
    fieldsV1:
      f:spec:
        f:replicas: {}
    manager: curl
    operation: Update
    subresource: scale
> ...

which lets the helm upgrade command fail. Did you experience the same issues?

maarous:
> Hi! When trying out your feature, my local cluster of 3 Keycloaks scales down, but not up anymore.
> 
> > `helm upgrade --install keycloak helm-charts/charts/keycloakx --values values.yaml --namespace pr898 --set image.tag=26.6.1`
> > `level=WARN msg="upgrade failed" name=keycloak error="conflict occurred while applying object pr898/keycloak-keycloakx apps/v1, Kind=StatefulSet: Apply failed with 1 conflict: conflict with \"curl\" with subresource \"scale\" using apps/v1: .spec.replicas"`
> > `Error: UPGRADE FAILED: conflict occurred while applying object pr898/keycloak-keycloakx apps/v1, Kind=StatefulSet: Apply failed with 1 conflict: conflict with "curl" with subresource "scale" using apps/v1: .spec.replicas `
> 
> The `curl` command modifies and takes ownership of the `scale` resource:
> 
> > ...
> > apiVersion: apps/v1
> > fieldsType: FieldsV1
> > fieldsV1:
> > f:spec:
> > f:replicas: {}
> > manager: curl
> > operation: Update
> > subresource: scale
> > ...
> 
> which lets the helm upgrade command fail. Did you experience the same issues?

Hello @dominiquemetz , thank you for your quick reply.

Unfortunately, I haven’t been able to reproduce the issue on my side.

I tested this setup locally using Minikube and also deployed it across our vanilla Kubernetes clusters in all environments without encountering any problems.

For reference, I retested using:
- Minikube v1.35
- Helm v3.19.4

The issue still did not occur.

Here is the values file I used:
```yaml
replicas: 3

command:
  - "/opt/keycloak/bin/kc.sh"
  - "start"
  - "--http-port=8080"
  - "--hostname-strict=false"
extraEnv: |
  - name: KEYCLOAK_ADMIN
    value: admin
  - name: KEYCLOAK_ADMIN_PASSWORD
    value: admin
  - name: JAVA_OPTS_APPEND
    value: >-
      -Djgroups.dns.query={{ include "keycloak.fullname" . }}-headless

affinity: ""

updateHook:
  enabled: true
  kubernetesApi:
    url: https://192.168.49.2
    port: 8443
    cidr: 192.168.49.2/32
```

I’ll attach the results here for reference.
[Screencast from 2026-05-01 13-54-04.webm](https://github.com/user-attachments/assets/de8f0e40-7350-406d-943b-0278c462aa0c)

Out of curiosity, are you using Helm v4?

maarous:
I think it's better to disable this feature by default (for new cluster set up and for patch update that are supported natively in Keycloak HA).

I will add also some improvements.

dominiquemetz:
Hi, sorry for the late response. I tested it more thoroughly and, yes, Helm v3 works without problems. My issues only come up with Helm v4. I did not find a solution for this issue yet. Do you have an idea how to get helm v4 working for this feature?

maarous:
Hi @dominiquemetz ,

After looking into this, I believe the most viable solution for Helm v4 is to use the `--force-conflicts` flag, which would force Helm to take ownership of the StatefulSet scale subresource.

I'm aware that this will make the change less transparent to other chart users but given the constraints, it may be the best (and possibly only) option here.

We could add a clear note in the documentation explaining the behavior and that this flag is needed with helm 4.

What do you think?

maarous:
Added a fix to prevent downscaling during patch updates, as Keycloak supports patch upgrades in HA mode.

dominiquemetz:
> Hi @dominiquemetz ,
> 
> After looking into this, I believe the most viable solution for Helm v4 is to use the `--force-conflicts` flag, which would force Helm to take ownership of the StatefulSet scale subresource.
> 
> I'm aware that this will make the change less transparent to other chart users but given the constraints, it may be the best (and possibly only) option here.
> 
> We could add a clear note in the documentation explaining the behavior and that this flag is needed with helm 4.
> 
> What do you think?

Using Helm v4, I could circumvent the issues with ownership by adding the flags `--force-conflicts=true` and ` --server-side=true`. I still had the issue that the scaling down by the pod worked (setting replicas to 1), but the rollingUpdate afterwards did NOT delete the last pod and scaled up.

Instead, it created a new pod with the new version, which got stuck due to JGroups incompatibility. Which is the issue we want to solve with this PR. Scaling down to 0 would fix this issue and is basically the `deployments` strategy `Recreate`. I do think this is a valid compromise when coupled with documentation about what is happening when using this functionality.

PR #901, opened 2026-05-06
fix(keycloakx): preserve YAML document separators in templates

### LINKED ISSUE #900: Helm lint fails when HTTPRoute and console HTTPRoute are enabled
When enabling `httpRoute` and `httpRoute.console`, `helm lint` fails with a YAML parsing error:

```text
templates/httproute.yaml: unable to parse YAML: invalid Yaml document separator: apiVersion: gateway.networking.k8s.io/v1
```

This appears to be caused by whitespace trimming around the second `HTTPRoute` document separator:

```yaml
---
{{- if $httpRoute.console.enabled -}}
apiVersion: gateway.networking.k8s.io/v1
```

Because both sides of the template directive trim whitespace, Helm can render this as:

```yaml
---apiVersion: gateway.networking.k8s.io/v1
```

which is invalid YAML.

Suggested fix: avoid double-sided trimming around the document separator, for example:

```yaml
---
{{- if $httpRoute.console.enabled }}
apiVersion: gateway.networking.k8s.io/v1
```

This should preserve the newline after `---` and make the rendered YAML valid.

Potentially other resources are problematic as well? I haven't checked every template for this issue

SeWieland:
Fixes #900 

I checked the whole repo and found only these two instances of this specific linting error.

### DIFF
--- a/charts/keycloakx/templates/httproute.yaml
+++ b/charts/keycloakx/templates/httproute.yaml
@@ -47,7 +47,7 @@ spec:
     {{- end }}
 {{- end }}
 ---
-{{- if $httpRoute.console.enabled -}}
+{{- if $httpRoute.console.enabled }}
 apiVersion: gateway.networking.k8s.io/v1
 kind: HTTPRoute
 metadata:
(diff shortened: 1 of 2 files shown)


dominiquemetz:
Looks good, thank you for the contribution!

PR #906, opened 2026-05-29
feature: add gatewayapi support in mailhog

pantovad:
Add support for Gateweay API in Mailhog chart

### DIFF
--- a/charts/mailhog/templates/gatewayapi_httproutes.yaml
+++ b/charts/mailhog/templates/gatewayapi_httproutes.yaml
@@ -1,0 +1,110 @@
+{{- if .Values.gatewayapi.enabled -}}
+{{- $fullName := include "mailhog.fullname" . -}}
+{{- if .Values.gatewayapi.tls }}
+---
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: {{ printf "%s-redirect" $fullName }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+    {{- include "mailhog.labels" . | nindent 4 }}
+    {{- with .Values.gatewayapi.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  parentRefs:
+    - group: gateway.networking.k8s.io
+      kind: ListenerSet
+      name: {{ $fullName }}
+      namespace: {{ .Release.Namespace | quote }}
+      sectionName: {{ printf "%s-http" $fullName }}
+  hostnames:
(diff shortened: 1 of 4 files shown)


charts/mailhog/templates/gatewayapi_httproutes.yaml
```diff
@@ -0,0 +1,110 @@
...
+  rules:
+    {{- range .Values.gatewayapi.hosts }}
+    {{- range .paths }}
+    - matches:
+        - path:
+            type: {{ .pathType | default "PathPrefix" }}
+            value: {{ .path }}
+      backendRefs:
+        - name: {{ $fullName }}
+          port: 8025
```
dominiquemetz, line 67: This should be {{ .Values.service.port.http }} right?

charts/mailhog/templates/gatewayapi_httproutes.yaml
```diff
@@ -0,0 +1,110 @@
...
```
dominiquemetz, line 106: This should be {{ .Values.service.port.http }} right?

charts/mailhog/values.yaml
```diff
@@ -71,14 +71,23 @@ ingress:
...
 
-
-  tls: []
-  #  - secretName: chart-example-tls
-  #    hosts:
-  #      - chart-example.local
+gatewayapi:
+  enabled: false
+  gateway:
+    name: envoy-gateway
```
dominiquemetz, line 80: We don't want to assume a specific controller. Please leave this empty

charts/mailhog/values.yaml
```diff
@@ -71,14 +71,23 @@ ingress:
...
-        - path: "/"
+        - path: /
           pathType: Prefix
 
-
-  tls: []
-  #  - secretName: chart-example-tls
-  #    hosts:
-  #      - chart-example.local
+gatewayapi:
```
dominiquemetz, line 77: Please provide an example for the tls block when using the Gateway for completions sake

dominiquemetz (changes requested):
Hi, thank you for the contribution!

The implementation looks good overall, and I've left some inline comments on specific lines. 

For the documentation, please update the README to cover the new `gatewayapi` section in `values.yaml`, including all fields and their purpose. Please also make sure to document that this feature uses `ListenerSet`, which as of June 2026 is only supported by Envoy Gateway v1.8+. No other Gateway controller implements it yet.

jullianow:
@pantovad Can you make the requested adjustments?
I would even suggest adding the new ListenerSet object right away: https://gateway-api.sigs.k8s.io/guides/user-guides/listener-set/

PR #908, opened 2026-06-26
feat(keycloakx): add namespaceOverride and service.internalTrafficPolicy

teddy0605:
## Summary

- Adds \`namespaceOverride\` value so umbrella charts can deploy multiple aliased keycloakx instances each into their own namespace. A \`keycloakx.namespace\` helper centralises the logic; all templates use it instead of \`.Release.Namespace\` directly.
- Adds \`service.internalTrafficPolicy\` to \`values.yaml\` and \`service-http.yaml\`. Previously there was no way to set this field without a post-renderer workaround.

## Motivation

When deploying keycloakx as a subchart multiple times in the same umbrella chart (using Helm aliases), all instances land in the release namespace. \`namespaceOverride\` lets each alias target a different namespace independently.

\`internalTrafficPolicy\` is needed to control routing for internal cluster traffic (e.g. \`Local\` to keep traffic node-local), which is a standard Kubernetes Service field not currently exposed by the chart.

## Backwards compatibility

Both changes are additive. Default values (\`namespaceOverride: ""\`, \`internalTrafficPolicy: ""\`) preserve existing behaviour — charts that don't set these values render identically to before.

### DIFF
--- a/charts/keycloakx/templates/NOTES.txt
+++ b/charts/keycloakx/templates/NOTES.txt
@@ -19,15 +19,15 @@ Keycloak was installed with a Service of type NodePort.
 {{ if .Values.service.httpNodePort }}
 Get its HTTP URL with the following commands:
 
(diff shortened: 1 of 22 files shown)


teddy0605:
Closing to reopen with a clean commit history.

PR #909, opened 2026-06-26
feat(keycloakx): add namespaceOverride and service.internalTrafficPolicy

teddy0605:
## Summary

- Adds `namespaceOverride` value so umbrella charts can deploy multiple aliased keycloakx instances each into their own namespace. A `keycloakx.namespace` helper centralises the logic; all templates use it instead of `.Release.Namespace` directly.
- Adds `service.internalTrafficPolicy` to `values.yaml` and `service-http.yaml`. Previously there was no way to set this field without a post-renderer workaround.

## Motivation

When deploying keycloakx as a subchart multiple times in the same umbrella chart (using Helm aliases), all instances land in the release namespace. `namespaceOverride` lets each alias target a different namespace independently.

`internalTrafficPolicy` is needed to control routing for internal cluster traffic (e.g. `Local` to keep traffic node-local), which is a standard Kubernetes Service field not currently exposed by the chart.

## Backwards compatibility

Both changes are additive. Default values (`namespaceOverride: ""`, `internalTrafficPolicy: ""`) preserve existing behaviour — charts that don't set these values render identically to before.


charts/keycloakx/templates/_helpers.tpl
```diff
@@ -62,7 +62,14 @@ Create the name of the service account to use
...
 */}}
 {{- define "keycloak.serviceDnsName" -}}
-{{ include "keycloak.fullname" . }}-headless.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
+{{ include "keycloak.fullname" . }}-headless.{{ include "keycloakx.namespace" . }}.svc.{{ .Values.clusterDomain }}
+{{- end }}
+
+{{/*
+Namespace for all resources. Callers can override via .Values.namespaceOverride.
+*/}}
+{{- define "keycloakx.namespace" -}}
```
dominiquemetz: Please rename it to "keycloak.namespace" to be consistent with the naming conventions we already have

charts/keycloakx/values.yaml
```diff
@@ -4,6 +4,10 @@ fullnameOverride: ""
 # Optionally override the name
 nameOverride: ""
 
+# Optionally override the namespace for all resources. Useful for umbrella charts that
+# deploy multiple aliased keycloakx instances each into their own namespace.
```
dominiquemetz: "keycloakx" -> "keycloak"

dominiquemetz (changes requested):
Looks great already! Two minor nitpicks about naming. Please also update the README.md, documenting the new properties you introduced

teddy0605:
> Looks great already! Two minor nitpicks about naming. Please also update the README.md, documenting the new properties you introduced

Thank you for the kind response. I have addressed all the comments.

dominiquemetz:
Perfect, thank you very much for your contribution!

PR #912, opened 2026-07-07
feat(keycloakx): add commonLabels to apply labels across all chart resources

### LINKED ISSUE #915: keycloakx: support commonLabels to apply labels across all chart resources
## Problem

The `keycloakx` chart has no single place to define a label once and have it applied to every resource. Applying the same label everywhere (e.g. a cost-allocation tag) currently requires copying it into ~13 separate per-resource `labels` fields (`podLabels`, `statefulsetLabels`, `service.*.labels`, `ingress.labels`, `route.labels`, `serviceMonitor.labels`, `prometheusRule.labels`, `hpa.labels`, etc.).

## Proposed solution

Add a `commonLabels` value that gets merged into every resource's labels, including the StatefulSet's pods, without touching selector fields (`matchLabels`) or any of the existing per-resource `labels` fields.

## PR

Implemented in #912

rbm897:
Closes #915

## Summary

Adds a `commonLabels` value to the `keycloakx` chart. Today, applying the same label to every resource in the chart (e.g. a cost-allocation tag) requires copying it into ~13 separate per-resource `labels` fields (`podLabels`, `statefulsetLabels`, `service.*.labels`, `ingress.labels`, etc.). `commonLabels` lets you define it once and have it applied everywhere, including the StatefulSet's pods.

## Changes

- `values.yaml` / `values.schema.json`: new `commonLabels: {}` value.
- `templates/_helpers.tpl`: `commonLabels` is merged into the existing `keycloak.labels` helper, which is already included by every resource template in the chart (StatefulSet, both Services, Ingress, HTTPRoute, Route, RBAC, ServiceAccount, ConfigMap, Secret, HPA, PodDisruptionBudget, PrometheusRule, ServiceMonitor, NetworkPolicy, test Pod/ConfigMap) — one change propagates to all of them.
- `templates/statefulset.yaml`: the pod template (`spec.template.metadata.labels`) doesn't include `keycloak.labels`, so `commonLabels` is applied there separately, between `selectorLabels` and `podLabels`.
- `README.md`: documents the new parameter.

Selector fields (`keycloak.selectorLabels` / all `matchLabels`) are untouched, and none of the existing per-resource `labels` fields are modified — `commonLabels` is purely additive.

On a label-key collision, the user's `commonLabels` value takes precedence over the chart's own built-in labels (`helm.sh/chart`, `app.kubernetes.io/version`, `app.kubernetes.io/managed-by`), and a resource's own specific `labels` field (e.g. `podLabels`) still takes precedence over `commonLabels`.

No manual chart version bump — this repo bumps automatically from commit message keywords (this PR's commit message includes `feat`, so it triggers a minor bump).

## Test plan

- [x] `helm lint ./charts/keycloakx` passes
- [x] `helm template` with `commonLabels` set renders the label on every resource kind the chart can produce (14 kinds with all optional resources enabled), plus once more on the StatefulSet's pod template
- [x] Collision precedence verified: a `commonLabels` key matching a built-in label (e.g. `app.kubernetes.io/managed-by`) ends up with the user's value
- [x] Collision precedence verified: a `commonLabels` key matching a `podLabels` key ends up with the `podLabels` value on the pod template
- [x] Default (`commonLabels: {}`) renders identically to the pre-change chart — no behavior change for existing users


charts/keycloakx/templates/_helpers.tpl
```diff
@@ -37,6 +37,9 @@ helm.sh/chart: {{ include "keycloak.chart" . }}
 {{ include "keycloak.selectorLabels" . }}
 app.kubernetes.io/version: {{ .Values.image.tag | default .Chart.AppVersion | toString | trunc 63 | quote }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- range $key, $value := .Values.commonLabels }}
+{{ printf "%s: %s" $key (tpl $value $ | quote) }}
```
dominiquemetz, line 41: 🟡 **Suggestion: Duplicate YAML keys on collision with built-in labels**

When a `commonLabels` key matches one of the built-in labels (e.g. `app.kubernetes.io/managed-by`), the rendered output contains **duplicate YAML keys**:

```yaml
labels:
  app.kubernetes.io/managed-by: "my-custom-tool"   # duplicate key
```

Kubernetes silently takes the last value (so `commonLabels` does win, as documented), but this is technically invalid YAML per the spec and can trip up strict validators or tooling that consumes the rendered manifests.

This is a very common pattern in Helm charts (Bitnami's common library does the same), and the PR description correctly documents the precedence, so it's acceptable as-is. If you ever want to avoid duplicate keys, the alternative would be to build a merged `dict` and render it once with `toYaml`, e.g.:

```helm
{{- $labels := dict "helm.sh/chart" (include "keycloak.chart" .) -}}
{{- $labels = mergeOverwrite $labels .Values.commonLabels -}}
{{- toYaml $labels -}}
```

No change required for this PR — just flagging for awareness.

charts/keycloakx/values.schema.json
```diff
@@ -109,6 +109,9 @@
     "command": {
       "type": "array"
     },
+    "commonLabels": {
+      "type": "object"
```
dominiquemetz, line 113: 🔵 **Suggestion: Consider enforcing string values**

Adding `"additionalProperties": {"type": "string"}` would give users a clear schema-validation error instead of a template crash if they accidentally pass a non-string value:

```
Error: wrong type for value; expected string; got float64
```

```json
"commonLabels": {
  "type": "object",
  "additionalProperties": { "type": "string" }
}
```

Note: this is a pre-existing pattern — `podLabels`, `statefulsetLabels`, and the other per-resource `labels` fields all have the same issue (`"type": "object"` without `additionalProperties`). So this isn't a regression, just an opportunity to improve the new value since it's the most broadly-applied one.

charts/keycloakx/templates/statefulset.yaml
```diff
@@ -38,6 +38,9 @@ spec:
         {{- end }}
       labels:
         {{- include "keycloak.selectorLabels" . | nindent 8 }}
+        {{- range $key, $value := .Values.commonLabels }}
+        {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 8 }}
```
dominiquemetz, line 42: 🔵 **Suggestion: Consider documenting the selector-label footgun**

The placement here (after `selectorLabels`, before `podLabels`) correctly preserves the documented precedence. However, if a user sets a `commonLabels` key that collides with a selector label (`app.kubernetes.io/name` or `app.kubernetes.io/instance`), the pod template labels will have the user's value while the StatefulSet's `matchLabels` selector keeps the chart's value — breaking selector matching and leaving the StatefulSet in a broken state.

This is the same footgun that already exists with `podLabels`, so it's not a regression. But since `commonLabels` is more broadly applied, it might be worth adding a note to the README or a `values.yaml` comment warning users not to override `app.kubernetes.io/name` or `app.kubernetes.io/instance`.

No code change needed — just a documentation thought.

dominiquemetz (commented):
Sorry, my llm agent has pushed its "findings" automatically as a review. I will need to verify the findings and will come back to this PR later

dominiquemetz:
Thank you for contribution, looks good to me!

rbm897:
Thank you for the review and merge.

PR #918, opened 2026-08-17
Keycloakx add missing trimsuffix

reixd:
Patched the keycloakx helm chart in order to avoid templating errors, if the docker image.tag is longer than 63 chars and after truncating the remaining string contains a dash "-" at the end.


dominiquemetz:
Looks good, thank you for the contribution!

PR #924, opened 2026-09-02
fix(keycloakx): HTTPRoute + ListenerSet honor namespaceOverride; console attaches to ListenerSet

MaxMuzh:
### TL;DR

Three small fixes in the Gateway API templates. Two are namespace mismatches when `namespaceOverride` is set; one is the console HTTPRoute not attaching to anything in a ListenerSet-only setup.

Fixes part of #926.

### The bugs

**1. `listenerset.yaml` — namespace ignores `namespaceOverride`.** The ListenerSet's own `metadata.namespace` uses `.Release.Namespace` directly. Every other template in the chart goes through `include "keycloak.namespace"`, which honors `.Values.namespaceOverride`. Set `namespaceOverride` and the ListenerSet lands in one namespace while the HTTPRoute lands in another — nothing attaches.

**2. `httproute.yaml` main route parentRef — same bug.** When `listenerSet.enabled: true`, the main HTTPRoute's `parentRef.namespace` also resolves to `.Release.Namespace` instead of the override.

**3. `httproute.yaml` console route ignores `listenerSet.enabled`.** The main route has a `listenerSet` branch that populates `parentRefs` and `hostnames` from the listeners. The console route doesn't — it only plucks from `console.parentRefs` / `httpRoute.parentRefs` and `console.hostnames` / `httpRoute.hostnames`. In a ListenerSet-only config (where neither `parentRefs` field is set), the console route renders empty `parentRefs` and attaches to no Gateway.

Fix mirrors the main route's `listenerSet` branch into the console route.

### Verified

`helm template` with `namespaceOverride=custom-ns` + `listenerSet.enabled: true` + `console.enabled: true`:
- `HTTPRoute` main → `namespace: custom-ns`, `parentRefs[0].namespace: custom-ns` ✅
- `HTTPRoute` console → `namespace: custom-ns`, `parentRefs[0].namespace: custom-ns`, `hostnames` populated ✅
- `XListenerSet` → `namespace: custom-ns` ✅

`helm lint` clean.

Companion to #925 (schema fix, unrelated file).

PR #925, opened 2026-09-02
fix(keycloakx): make top-level values.schema.json keys actually validate

MaxMuzh:
### TL;DR

About 40 top-level keys in `values.schema.json` sit at the wrong nesting depth, so they're silently ignored by JSON Schema. `replicas: "three"` passes today. This moves them back to the top-level `properties` map.

Fixes part of #926.

### What's going on

The keys ended up as siblings of `ingress.properties` inside the `ingress` object, instead of at the top-level `properties`. Since JSON Schema only validates keys declared inside a `properties` object, ~40 keys (`replicas`, `tolerations`, `resources`, `namespaceOverride`, `serviceAccount`, and so on) are treated as unknown keywords and validate nothing.

Two related things needed fixing at the same time:

- The `http` block uses literal strings like `"relativePath": "string"` instead of `{"type": "string"}` sub-schemas — same silent-ignore issue, tiny fix.
- Once the schema starts validating, two existing chart defaults trip it: `dbchecker.image.tag: 1.32` (YAML parses `1.32` as a float, so `tag` needs `number`), and `database.*` fields (null by default, need `null` alongside their real type). I loosened those types so `helm lint` passes on defaults.

### On the diff size — sorry, this one is unavoidable

Moving the 38 misplaced keys out of `ingress` means dedenting them by 2 spaces. `git diff` counts that as delete+add per line, so the raw diff is ~700 lines. **`git diff -w` shows the real change: 16 additions, 13 deletions.** No formatting reshuffling beyond the required dedent.

### Verified

- `helm lint` clean on chart defaults.
- `helm template` with `namespaceOverride=custom-ns`, `replicas=3`, and a `tolerations` array all render fine.
- Attempting `replicas: "three"` now surfaces a real schema error (previously silent).

Companion to #924 — kept separate because these edits are all in `values.schema.json` and can be reviewed on their own.

### DIFF
--- a/charts/keycloakx/values.schema.json
+++ b/charts/keycloakx/values.schema.json
@@ -77,7 +77,7 @@
           "type": "string"
         },
         "tag": {
-          "type": ["string", "integer"]
+          "type": ["string", "integer", "number"]
         },
         "digest": {
           "type": "string"
@@ -146,10 +146,13 @@
       "type": "array"
     },
     "http": {
-      "relativePath": "string",
-      "managementRelativePath": "string",
-      "internalPort": "string",
-      "internalScheme": "string"
+      "type": "object",
+      "properties": {
+        "relativePath": { "type": "string" },
+        "managementRelativePath": { "type": "string" },
+        "internalPort": { "type": ["string", "integer"] },
+        "internalScheme": { "type": "string" }
+      }
     },
     "httpRoute": {
       "allOf": [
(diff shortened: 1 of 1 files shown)


MaxMuzh:
@codecentric/helm-charts-contributors — friendly ping. Schema restructure so top-level keys actually validate (currently silently ignored). Diff looks large but that's the dedent tax — `git diff -w` shows the real change is 16 additions, 13 deletions. CI + DCO green. Would love a review. Thanks!

MaxMuzh:
Correction to my previous ping — the `@codecentric/helm-charts-contributors` team handle isn't resolving. @dominiquemetz @stefangries @jenslehnhoff — you've been actively merging kcx PRs, would appreciate a look at this one when you have a moment. Schema restructure so top-level keys actually validate. Diff looks large but `git diff -w` shows the real change is 16 additions, 13 deletions. CI + DCO green. Thanks!

stefangries:
Hi @MaxMuzh! Thank you for your contribution. We'll review your pull request, but we can't promise we'll get to it this week—please give us a few days.

MaxMuzh:
@stefangries @dominiquemetz @jenslehnhoff - friendly nudge on #925. 😇
DCO + triage green, mergeable. 
#924 is the sibling PR in the same area if you'd like to bundle. 
Happy to rebase either. Thanks!🙏

Issue #917, opened 2026-07-16
[keycloakx] extraInitContainers & other extra* values should accept structured YAML lists, not only tpl strings


gecube:
## Summary

In the keycloakx chart, six `extra*` values are typed as plain **strings** rendered through `tpl`:

- [`values.yaml`](https://github.com/codecentric/helm-charts/blob/master/charts/keycloakx/values.yaml): `extraInitContainers`, `extraContainers`, `extraEnv`, `extraEnvFrom`, `extraVolumes`, `extraVolumeMounts` — all default to `""`
- rendered via `{{- tpl . $ | nindent 8 }}` (e.g. [`statefulset.yaml#L71-L73`](https://github.com/codecentric/helm-charts/blob/master/charts/keycloakx/templates/statefulset.yaml#L71-L73))
- [`values.schema.json`](https://github.com/codecentric/helm-charts/blob/master/charts/keycloakx/values.schema.json) enforces `"type": "string"` for all six, so passing a normal YAML list fails hard at install time with a schema validation error.

Meanwhile the chart is inconsistent with itself: `extraPorts` in the same schema is already `"type": "array"`, and the README documents `extraInitContainers` / `extraContainers` as `[]`. That documentation mismatch was reported in #862 and got closed by the stale bot without an answer — this issue is the follow-up, with the concrete breakage that string-typed manifests cause.

## Why string-typed manifests are a problem

Kubernetes manifests embedded in a block scalar are opaque to every tool that processes values as YAML. A concrete, real-world case I just finished triaging: **fluxcd/flux2#5307**. Flux image automation updates image tags by looking for `# {"$imagepolicy": ...}` marker comments attached to YAML field nodes. When an init container lives inside `extraInitContainers: |`, the marker is string *content*, not a YAML comment — the automation silently skips it, and the user is left wondering why their theme-provider image never gets bumped. Full analysis with a reproduction: https://github.com/fluxcd/flux2/issues/5307#issuecomment-4988860477

The same opacity breaks Renovate's `helm-values` manager, structural patching of HelmRelease values (kustomize / JSON6902), and schema-aware editors — plus the everyday papercut that there is no YAML validation or completion inside a string block, so typos surface only at render time.

## Proposal

Accept **both** types — keep strings working for backward compatibility, additionally allow structured lists. This is the standard pattern used across the ecosystem (grafana charts, bitnami `common.tplvalues.render`):

```yaml
{{- with .Values.extraInitContainers }}
[...]

PR #927, opened 2026-09-16
fix(keycloakx): allow extraEnv to accept a structured YAML list

MaxMuzh:
### TL;DR

`extraEnv` is typed as `string` and rendered through `tpl`, so manifests embedded in a block scalar are opaque to Flux image automation, Renovate, structural patchers, and schema-aware editors. This loosens the schema to `oneOf: [string, array]` and branches the template on `kindIs "string"`. String form is unchanged.

Fixes part of #917.

### The bug

`extraEnv` in `values.schema.json` is `{"type": "string"}` and rendered in `templates/statefulset.yaml` via `{{- tpl . $ | nindent 12 }}`. Passing a normal YAML list fails schema validation. The block-scalar workaround hides field content from any tool that walks values as YAML:

- Flux image automation silently skips `# {"$imagepolicy": ...}` markers inside string content — full repro: fluxcd/flux2#5307
- Renovate's `helm-values` manager can't see fields inside the scalar
- Structural patching (JSON6902 / kustomize) can't reach them
- Schema-aware editors give no completion or validation

The chart is already inconsistent with itself — `extraPorts` in the same schema is `type: array`.

### The fix

- `values.schema.json`: `extraEnv` → `oneOf: [{type: string}, {type: array}]`.
- `templates/statefulset.yaml`: branch on `kindIs "string"`.
  - string → `{{- tpl . $ | nindent 12 }}` (existing path)
  - array  → `{{- tpl (toYaml .) $ | nindent 12 }}` — the `tpl` wrap keeps `{{ .Release.Namespace }}` and other template expressions working inside array values, so array mode isn't a downgrade.
- `README.md`: one-line note on the parameter row.

### Verified

`helm template` with `--namespace demo-ns` in both modes:

- **String mode** (existing behavior):
  ```
  - name: FOO_STRING
    value: "bar"
  - name: NS_STRING
    value: "demo-ns"
  ```
- **Array mode** (new):
  ```
  - name: JAVA_OPTS_APPEND
    value: -Djgroups.dns.query=keycloak-headless.demo-ns.svc.cluster.local
  ```
  `{{ .Release.Namespace }}` interpolates correctly inside array values.

All existing values files render byte-identically. `helm lint` clean.

### Scope

`extraEnv` only. The same treatment is needed on `extraContainers`, `extraInitContainers`, `extraEnvFrom`, `extraVolumes`, `extraVolumeMounts` — I'll open a follow-up PR for those once this pattern is reviewed.


MaxMuzh:
@stefangries @dominiquemetz — friendly nudge, and thanks for the recent merge cadence on #924 and #925. Same pattern here: schema loosens `extraEnv` from string-only to `oneOf: [string, array]` and the template branches on `kindIs "string"`. String form is byte-identical, verified with `helm template` on both shapes. DCO + triage green, mergeable. Small diff. Would love to get this in the next chart cut if you have a moment.

